PR #3, opened 2020-04-02
Add line item validation

jtmckibb:
- Should be greater than $.50
 - Should be a gift card or donation
 - Should have all required attributes:
   amount, currency, name, and quantity

### DIFF
--- a/api/app/controllers/charges_controller.rb
+++ b/api/app/controllers/charges_controller.rb
@@ -1,5 +1,6 @@
 
 class ChargesController < ApplicationController
+  class InvalidLineItem < StandardError; end
 
   # POST /charges
   def create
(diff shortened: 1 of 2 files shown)


api/app/controllers/charges_controller.rb
```diff
@@ -41,4 +44,19 @@ def charge_params
     params.require(:line_items)
     params.permit(:merchant_id, line_items: [[:amount, :currency, :name, :quantity, :description]])
   end
+
+  def validate(line_item:)
+    [:amount, :currency, :name, :quantity].each do |attribute|
+      unless line_item.key?(attribute)
+        raise ActionController::ParameterMissing.new "param is missing or the value is empty: #{attribute}"
+      end
+    end
```
PhilipPachigalla, line 53:
```suggestion
unless ([:amount, :currency, :name, :quantity] - line_item.keys).empty?
  raise ActionController::ParameterMissing.new "all of a,b,c,d needed" 
end
    end
```

You do miss out by not seeing which is missing, but in the former, you find out one by one 🤷‍♂

api/app/controllers/charges_controller.rb
```diff
@@ -41,4 +44,19 @@ def charge_params
...
+
+    unless ['Gift Card', 'Donation'].include? line_item['name']
+      raise InvalidLineItem.new "line_item must be named `Gift Card` or `Donation`"
+    end
+
+    amount = line_item['amount'].to_i
```
PhilipPachigalla, line 59: `to_i` won't throw if it's not an integer, so if the front end starts sending you dollar amounts with decimals, this won't catch it

api/app/controllers/charges_controller.rb
```diff
@@ -29,6 +30,8 @@ def create
       json_response(e.error.message, e.http_status)
     rescue ActionController::ParameterMissing => e
       json_response(e.message, :unprocessable_entity)
+    rescue InvalidLineItem => e
```
PhilipPachigalla, line 33: You should be able to do:
```
rescue ActionController::ParameterMissing, InvalidLineItem => e
```

PR #10, opened 2020-04-06
adding cors

stanleyzhengnyc:
enables us to make requests from our from our main domain and local/dev


config/application.rb
```diff
@@ -24,6 +24,17 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
```
stanleyzhengnyc, line 24: Q about CORS

PR #13, opened 2020-04-09
Add new webhook for Stripe

jtmckibb:
Delete old create gift cards endpoint
Use new Item model
Create new GiftCardDetail Model
Create new GiftCardAmount Model
Use new DontationDetail Model
Disable GiftCard endpoint tests because
we aren't using them yet


app/controllers/webhooks_controller.rb
```diff
@@ -0,0 +1,85 @@
+class WebhooksController < ApplicationController
+  # POST /webhooks
+  def create
+    payload = request.body.read
+    event = nil
+
+    # Verify webhook signature and extract the event
+    # See https://stripe.com/docs/webhooks/signatures for more information.
+    sig_header = request.env['HTTP_STRIPE_SIGNATURE']
+    endpoint_secret = 'whsec_8I8JI2kT6B3RdUR39usVosRMhRvLVrcH'
```
stanleyzhengnyc, line 10: we should use ruby secrets store more credentials like these

app/models/gift_card_detail.rb
```diff
@@ -0,0 +1,5 @@
+class GiftCardDetail < ApplicationRecord
+  belongs_to :item
+  validates_uniqueness_of :gift_card_id
+  has_many :gift_card_amount
```
stanleyzhengnyc, line 4: a gift card has multiple amounts?  is this to cover spending on the amt? 

just purely a question

spec/requests/gift_cards_spec.rb
```diff
@@ -3,10 +3,11 @@
 RSpec.describe 'Gift Cards API', type: :request do
 
   # initialize test data
-  let!(:gift_card) { create(:gift_card) }
```
stanleyzhengnyc, line 6: cool we have some tests; we could run some CI before anything gets meged

app/models/gift_card_detail.rb
jtmckibb: This is to track the history of the spending of a gift card, otherwise it will be difficult to reconcile reporting, undoing an accidental charge, etc.

app/controllers/webhooks_controller.rb
jtmckibb: Agreed. This is a test secret so it's nbd right now. But we should pull all of these out and use ruby secrets and heroku config vars to deploy

PR #25, opened 2020-04-17
Fix the Stripe webhook and amount_raised

jtmckibb:
The Stripe webhook was broken because it wasn't
the Stripe API key for header verification.

Also, amount_raised was broken because I was
using the wrong seller_id. I was passing in
Seller.seller_id when I should have been
using Seller.id. I added an integration test
to cover this.

Finally I updated the factory for Seller because
using a single "word" was causing a lot of
collisions in the test db.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
 web: bundle exec puma -t 5:5 -p ${PORT:-3000} -e ${RACK_ENV:-development}
+log: tail -f log/development.log
--- a/app/controllers/concerns/exception_handler.rb
+++ b/app/controllers/concerns/exception_handler.rb
@@ -31,7 +31,7 @@ module ExceptionHandler
 
     # Invalid signature
     rescue_from Stripe::SignatureVerificationError do |e|
-      json_response({ message: e.error.message }, e.http_status)
+      json_response({ message: e.message }, e.http_status)
     end
 
   end
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -1,14 +1,13 @@
 class WebhooksController < ApplicationController
   # POST /webhooks
   def create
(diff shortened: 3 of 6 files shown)


Procfile
```diff
@@ -1 +1,2 @@
-web: bundle exec puma -t 5:5 -p ${PORT:-3000} -e ${RACK_ENV:-development}
\ No newline at end of file
+web: bundle exec puma -t 5:5 -p ${PORT:-3000} -e ${RACK_ENV:-development}
+log: tail -f log/development.log
```
jtmckibb, line 2: I added this so that heroku run web shows errors as they happen

app/controllers/concerns/exception_handler.rb
```diff
@@ -31,7 +31,7 @@ class CannotGenerateUniqueHash < StandardError; end
 
     # Invalid signature
     rescue_from Stripe::SignatureVerificationError do |e|
-      json_response({ message: e.error.message }, e.http_status)
+      json_response({ message: e.message }, e.http_status)
```
jtmckibb, line 34: This wasn't correctly printing the error because e.error was nil

app/controllers/webhooks_controller.rb
```diff
@@ -1,14 +1,14 @@
...
-    event = nil
+    Stripe.api_key = ENV['STRIPE_API_KEY']
 
     # Verify webhook signature and extract the event
     # See https://stripe.com/docs/webhooks/signatures for more information.
     sig_header = request.env['HTTP_STRIPE_SIGNATURE']
-
     endpoint_secret = ENV['STRIPE_WEBHOOK_KEY']
+    payload = request.body.read
+    event = nil
```
Yong-L, line 11: minor - is this necessary since event is assigned right below?

PR #27, opened 2020-04-17
Add MenuItems Controller

Yong-L:
Adds basic MenuItemsController as per https://docs.google.com/document/d/1DZDoGZPtpxOujtzKQutuXMFxtPd-gjsHsvbnTvqSvCM/edit#

### DIFF
--- a/app/controllers/menu_items_controller.rb
+++ b/app/controllers/menu_items_controller.rb
@@ -9,23 +9,32 @@ class MenuItemsController < ApplicationController
 
   # POST /sellers/:seller_id/menu_items
   def create
-    json_response(@seller.menu_items.create!(menu_item_params), :created)
+    json_response(@seller.menu_items.create!(create_menu_item_params), :created)
   end
 
   # PUT /sellers/:seller_id/menu_items/:id
   def update
-    @menu_item.update(menu_item_params)
+    @menu_item.update(update_menu_item_params)
     json_response(@menu_item)
   end
 
   # DELETE /sellers/:seller_id/menu_items/:id
   def destroy
-    json_response(@menu_item.destroy)
+    @menu_item.destroy
+
+    head :no_content
   end
 
   private
 
-  def menu_item_params
+  def create_menu_item_params
+    params.require(:seller_id)
+    update_params
+  end
+
+  def update_menu_item_params
+    params.require(:seller_id)
+    params.require(:id)
     update_params
   end
 
--- a/app/controllers/menu_items_controller.rb
+++ b/app/controllers/menu_items_controller.rb
@@ -1,0 +1,48 @@
+class MenuItemsController < ApplicationController
+  before_action :set_seller
+  before_action :set_seller_menu_item, only: [:update, :destroy]
+
+  # GET /sellers/:seller_id/menu_items
+  def index
+    json_response(@seller.menu_items)
+  end
+
+  # POST /sellers/:seller_id/menu_items
+  def create
+    json_response(@seller.menu_items.create!(menu_item_params), :created)
+  end
+
+  # PUT /sellers/:seller_id/menu_items/:id
+  def update
+    @menu_item.update(menu_item_params)
+    json_response(@menu_item)
+  end
+
+  # DELETE /sellers/:seller_id/menu_items/:id
+  def destroy
+    json_response(@menu_item.destroy)
+  end
+
+  private
+
+  def menu_item_params
+    update_params
+  end
+
+  def update_params
+    params.permit(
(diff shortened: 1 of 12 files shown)


app/controllers/menu_items_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  # DELETE /sellers/:seller_id/menu_items/:id
+  def destroy
+    json_response(@menu_item.destroy)
+  end
+
+  private
+
+  def menu_item_params
+    update_params
+  end
```
Yong-L, line 30: Kept this here for consistency with the other controllers

app/controllers/menu_items_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  # PUT /sellers/:seller_id/menu_items/:id
+  def update
+    @menu_item.update(menu_item_params)
+    json_response(@menu_item)
+  end
+
+  # DELETE /sellers/:seller_id/menu_items/:id
+  def destroy
+    json_response(@menu_item.destroy)
```
jtmckibb, line 23: this should be head :no_content

app/controllers/menu_items_controller.rb
```diff
@@ -0,0 +1,48 @@
+class MenuItemsController < ApplicationController
```
jtmckibb, line 1: I know that I don't do this, but can we add a MenuItemsControllerSpec so that we can test that the required params are actually required, and that the GET returns all the menu_items, POST creates one, etc.

If you look at request/sellers_spec.rb, I've been using the requests spec as an integration test and it was a bad idea because we end up with a super test file. It would be better to unit test each controller

app/controllers/menu_items_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    json_response(@menu_item.destroy)
+  end
+
+  private
+
+  def menu_item_params
+    update_params
+  end
+
+  def update_params
```
jtmckibb, line 32: Also be sure to test that if you pass in invalid params, that it filters them out

spec/requests/menu_items_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+  end
+
+  # Test suite for POST /sellers/:seller_id/menu_items
+  describe 'POST /sellers/:seller_id/menu_items' do
+    let(:valid_attributes) do
+      {
+        name: 'Food',
+        description: 'Awesome Food',
+        amount: 15,
```
jtmckibb, line 43: FYI the amount should be in cents

spec/controllers/menu_items_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe MenuItemsController, type: :controller do
```
jtmckibb, line 3: Oh I found the test

spec/models/menu_item_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe MenuItem, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
jtmckibb, line 4: Could you test 
   ` it { should belong_to(:seller) }`

spec/requests/menu_items_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+        image_url: 'image.com'
+      }
+    end
+
+    context 'when request attributes are valid' do
+      before { post "/sellers/#{seller_id}/menu_items", params: valid_attributes, as: :json }
+
```
jtmckibb, line 50: Could you test that it creates the new menu item? aka Seller.menu_items is eq to what you expect?

spec/requests/menu_items_spec.rb
```diff
@@ -0,0 +1,128 @@
+require 'rails_helper'
+
+RSpec.describe 'MenuItems API' do
```
jtmckibb, line 3: Oh wait I noticed you didn't test DELETE

Yong-L: Ah good catch

jtmckibb: Something like:

`# Test suite for DELETE /sellers/:id
  describe 'DELETE /sellers/:id' do
    before { delete "/sellers/#{seller_id}/menu_items/#{id}" }

    it 'deletes the seller' do
       ...
    end 

    it 'returns status code 204' do
      expect(response).to have_http_status(204)
    end
  end`

PR #29, opened 2020-04-18
add extra info fields to seller model

juliexxia:
Adding a flexible model that will be attached to sellers with all-optional fields for extra info like website and # employees

image


.gitignore
```diff
@@ -28,3 +28,7 @@
 
 # Ignore local environment variables
 .env
+
+# Ignore intellij stuff
```
jtmckibb, line 32: nice

spec/models/seller_extra_info_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe SellerExtraInfo, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
jtmckibb, line 4: Could you add a test:
    it { should belong_to(:seller) }

db/schema.rb
```diff
@@ -75,6 +75,16 @@
     t.boolean "successful", default: false
   end
 
+  create_table "seller_extra_infos", force: :cascade do |t|
+    t.string "type"
+    t.integer "num_employees"
+    t.integer "founded"
```
jtmckibb, line 81: Could we update this to founded_year?

app/models/seller_extra_info.rb
```diff
@@ -0,0 +1,3 @@
+class SellerExtraInfo < ApplicationRecord
```
jtmckibb, line 1: Mind adding a verify that founded is greater than like 1800 and <= the current year?

jtmckibb: or whatever you think is a good floor for merchants

spec/models/seller_extra_info_spec.rb
juliexxia: Can I do that at this point even though I haven't migrated sellers to know about SellerExtraInfos?

I added this one line in and the test failed though that might totally be because I'm missing something else

app/models/seller_extra_info.rb
juliexxia: Done. Any chance we have a place founded next year/in the future that ends up onboarding onto this? I guess unlikely.

.gitignore
juliexxia: (:

db/schema.rb
juliexxia: done

spec/models/seller_extra_info_spec.rb
jtmckibb: Ah you have to add subject { create :seller_extra_info } to use the SellerExtraInfo factory to generate a random object with an associated random seller

app/models/seller_extra_info.rb
jtmckibb: Probably unlikely

juliexxia:
Wasn't sure what would happen if I added the reference without running the migration? So just ran the migration for good measure.

app/models/seller.rb
```diff
@@ -2,6 +2,7 @@ class Seller < ApplicationRecord
   # model association
   has_many :locations, dependent: :destroy
 
+  validates :founded_year, :inclusion => 1800..2020
```
jtmckibb, line 5: could you test this in the model/seller_spec

.gitignore
```diff
@@ -28,3 +28,7 @@
 
 # Ignore local environment variables
 .env
```
jtmckibb: can you rebase onto master? I think I deleted this

jtmckibb: or if I didn't, mind deleting this? I checked in .env

spec/models/seller_spec.rb
```diff
@@ -5,15 +5,55 @@
   # ensure Seller model has a 1:m relationship with the MenuItem model
   # it { should have_many(:menu_items).dependent(:destroy) }
   # Validation tests
-  let!(:seller) { Seller.create(seller_id: 'oiawjefoiwjaef') }
+  # let!(:seller) { create :seller }
```
jtmckibb, line 8: might should delete

spec/models/seller_spec.rb
```diff
@@ -5,15 +5,55 @@
...
+
+  before { create :seller}
```
jtmckibb, line 10: nit: space after seller

spec/models/seller_spec.rb
```diff
@@ -5,15 +5,55 @@
...
   it { should validate_uniqueness_of(:seller_id) }
   it do
     should allow_value(%w[true false]).for(:sell_gift_cards)
   end
   it do
     should allow_value(%w[true false]).for(:accept_donations)
   end
+
+  let(:seller) do
```
jtmckibb, line 19: I think you can actually do let! and delete the before. ! happens before the before. I could be wrong here

spec/models/seller_spec.rb
```diff
@@ -5,15 +5,55 @@
...
+    create(:seller)
+  end
   it { expect(seller.target_amount).to eq 1_000_000 }
-  it { expect(seller.accept_donations).to eq true }
-  it { expect(seller.sell_gift_cards).to eq false }
+
+  # test founding year constraints
+  let(:time_travelling_seller) do
+    create(
```
jtmckibb, line 26: Oh do you mind setting your intellij settings to be spaces = 2 and not tabs?

jtmckibb: lol it's tabs vs spaces talk ahaha

spec/models/seller_spec.rb
```diff
@@ -5,15 +5,55 @@
...
+        :seller,
+        founded_year: founded_year,
+        )
+  end
+
+  context "test pre-1800 founding year" do
+    let(:founded_year) { '1' }
```
jtmckibb, line 33: LOL year 1

jtmckibb: This should be an int right?

spec/requests/sellers_spec.rb
```diff
@@ -27,7 +32,12 @@
       owner_name: 'Tom Hanks',
       owner_image_url: 'https://www.aws.com/oawjeoiajwef9wuef09wuef09waeuf',
       accept_donations: false,
-      sell_gift_cards: true
+      sell_gift_cards: true,
+      business_type: "medium-biz",
+      num_employees: 10,
+      founded_year: 1950,
+      website_url: 'https://www.youtube.com/watch?v=C_oACPWGvM4',
+      menu_url: 'https://www.youtube.com/watch?v=C_oACPWGvM4',
```
jtmckibb, line 40: to fix this, we could use the factory. You don't have to do this though

PR #30, opened 2020-04-18
Add Square Payment id

jtmckibb:


db/schema.rb
```diff
@@ -73,6 +73,7 @@
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.boolean "successful", default: false
+    t.string "square_payment_id"
```
juliexxia, line 76: Is there a way to verify that either square_payment_id != || stripe_id != null but not both?

jtmckibb: Yeah there is!

PR #31, opened 2020-04-18
Update charge create to be able to use square

juliexxia:


app/controllers/charges_controller.rb
```diff
@@ -24,31 +25,57 @@ def create
...
-      email: charge_params[:email],
-      line_items: line_items.to_json
-    )
-
-    json_response(intent)
+    if ENV['use_square']
+      api_client = Square::Client.new(
+        # TODO: put in ENV
+        access_token: 'EAAAEAQN-8I264d9ntPRq8iqR2I9jY94VEmdngHcGAghHSdAXjdFVcp7-DfECWXI',
+        # environment: ENV['IS_PRODUCTION'] == 'false' ? 'sandbox' : 'production'
```
jtmckibb, line 33: everything in the .env file is test/dev. production variables are stored on heroku

app/controllers/charges_controller.rb
```diff
@@ -24,31 +25,66 @@ def create
...
+      )
+
+      request_body = {
+        source_id: charge_params[:nonce],
+        idempotency_key: SecureRandom.uuid,
```
jtmckibb, line 38: Is this from Square? Do we need to store this id?

app/controllers/charges_controller.rb
```diff
@@ -24,31 +25,66 @@ def create
...
-    # Creates a pending PaymentIntent. See webhooks_controller to see what happens
-    # when the PaymentIntent is successful.
-    PaymentIntent.create!(
-      stripe_id: intent['id'],
-      email: charge_params[:email],
-      line_items: line_items.to_json
-    )
-
-    json_response(intent)
+    if ENV['use_square']
```
jtmckibb, line 29: Should be uppercase and have .present?

app/controllers/charges_controller.rb
```diff
@@ -24,31 +25,66 @@ def create
...
+      # when the PaymentIntent is successful.
+      # TODO: update stripe_id col to something payment agnostic
+      PaymentIntent.create!(
+          stripe_id: payment['id'],
+          email: charge_params[:email],
+          line_items: line_items.to_json
+      )
+
+      json_response(payment)
+    else
```
jtmckibb, line 58: Could we put this into a method?

app/controllers/charges_controller.rb
```diff
@@ -24,31 +25,66 @@ def create
...
+
+      intent = Stripe::PaymentIntent.create(
+        amount: amount,
+        currency: 'usd',
+        receipt_email: charge_params[:email],
+        payment_method_types: ['card'],
+        description: description
+      )
+
+      # Creates a pending PaymentIntent. See webhooks_controller to see what happens
```
jtmckibb, line 69: this part of the method is shared by both

app/controllers/charges_controller.rb
```diff
@@ -21,39 +20,40 @@ def create
...
     params.require(:line_items)
     params.require(:email)
-    params.permit(:email, line_items: [[:amount, :currency, :item_type, :quantity, :seller_id]])
+    params.permit(:email, :nonce, line_items: [[:amount, :currency, :item_type, :quantity, :seller_id]])
   end
 
   def validate(line_item:)
     [:amount, :currency, :item_type, :quantity, :seller_id].each do |attribute|
-    unless line_item.key?(attribute)
+      unless line_item.key?(attribute)
```
jtmckibb, line 56: I think this tab is supposed to be here

app/controllers/charges_controller.rb
```diff
@@ -21,39 +20,40 @@ def create
...
+    json_response(payment)
   end
 
   private
 
   def charge_params
     params.require(:line_items)
     params.require(:email)
-    params.permit(:email, line_items: [[:amount, :currency, :item_type, :quantity, :seller_id]])
+    params.permit(:email, :nonce, line_items: [[:amount, :currency, :item_type, :quantity, :seller_id]])
```
jtmckibb, line 51: nonce should be required for Square payments huh?

app/controllers/charges_controller.rb
```diff
@@ -91,4 +91,55 @@ def generate_description(seller_names:, item_types:)
...
+
+    request_body = {
+        source_id: source_id,
+        idempotency_key: SecureRandom.uuid,
+        amount_money: {
+            amount: amount,
+            currency: 'USD',
+        },
+        buyer_email_address: email,
+        note: note,
```
jtmckibb, line 108: Do you know if note gets emailed to the buyer?

app/controllers/charges_controller.rb
jtmckibb: If so, I'm starting to think that we should let the frontend pass a param that indicates whether or not they're paying with Square or Stripe, since it ultimately is on them

jtmckibb:
Looks good! Just a few nits. The whole frontend stripe/square thing we can talk about in a follow up

app/controllers/charges_controller.rb
juliexxia: Sorry did you mean you don't think the tab is supposed to be here? I think my auto-formatter did this

app/controllers/charges_controller.rb
juliexxia: I saw another way but it also seems to use a nonce: https://developer.squareup.com/docs/payments-api/take-payments#take-a-payment-using-a-card-on-file

I hear you - the param sounds good

app/controllers/charges_controller.rb
juliexxia: docs say "An optional note to be entered by the developer when creating a payment

Limit 500 characters."

so seems like no - so we can add that in later if we'd like! I'll add a tentative TODO

jtmckibb: Sounds good! Yeah we need it to appear on the receipt. I have a feeling we might have to create an "Order"

PR #53, opened 2020-04-24
start adding email receipt functionality

juliexxia:
HTML for email is very bare bones - I figure we can spruce it up in a follow up


db/migrate/20200424004105_add_receipt_url_to_payment_intents.rb
```diff
@@ -0,0 +1,5 @@
+class AddReceiptUrlToPaymentIntents < ActiveRecord::Migration[6.0]
+  def change
+    add_column :payment_intents, :receipt_url, :string
```
jtmckibb, line 3: remember to update the factory too

juliexxia: Thanks!

PR #54, opened 2020-04-24
Add name param to charge post action

juliexxia:


jtmckibb:
I have a feeling that test on the `sellers_helper_spec` failed though

juliexxia:
Good call! fixed

PR #60, opened 2020-04-25
Added existing item model to dedeup donations/giftcards/refunds

danthaman44:
1. Created new DB migration for existing_items table to track duplicate transactions
2. Created ActiveRecord model for ExistingItem

### DIFF
--- a/app/models/existing_event.rb
+++ b/app/models/existing_event.rb
@@ -1,0 +1,5 @@
+class ExistingEvent < ApplicationRecord
+  validates_presence_of :idempotency_key, :type
+  validates_uniqueness_of :idempotency_key
+  enum type: [:charges_create, :webhooks_create_refund_updated]
+end
--- a/db/migrate/20200425163638_create_existing_event.rb
+++ b/db/migrate/20200425163638_create_existing_event.rb
@@ -1,0 +1,8 @@
+class CreateExistingEvent < ActiveRecord::Migration[6.0]
+  def change
+    create_table :existing_events do |t|
+      t.string :idempotency_key
+      t.integer :type
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_04_25_034235) do
+ActiveRecord::Schema.define(version: 2020_04_25_163638) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -21,6 +21,11 @@ ActiveRecord::Schema.define(version: 2020_04_25_034235) do
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.index ["item_id"], name: "index_donation_details_on_item_id"
+  end
+
+  create_table "existing_events", force: :cascade do |t|
+    t.string "idempotency_key"
+    t.integer "type"
   end
 
   create_table "gift_card_amounts", force: :cascade do |t|
--- a/spec/factories/existing_event.rb
+++ b/spec/factories/existing_event.rb
@@ -1,0 +1,13 @@
+FactoryBot.define do
+  factory :existing_event do
+    idempotent_key { Faker::Alphanumeric.alphanumeric(number: 64) }
+
+    trait :charge do
+      type { :charges_create }
+    end
+
+    trait :webhook do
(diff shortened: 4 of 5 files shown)


app/models/existing_item.rb
```diff
@@ -0,0 +1,5 @@
+class ExistingItem < ApplicationRecord
+  validates_presence_of :idempotent_key, :item_type
```
jtmckibb, line 2: nit, can you name it `idempotency_key`

db/migrate/20200425033724_create_existing_items.rb
```diff
@@ -0,0 +1,8 @@
+class CreateExistingItems < ActiveRecord::Migration[6.0]
```
danthaman44, line 1: Should this be named "ExistingTransaction"?

jtmckibb: I'd generalize it and call it `Event`

db/migrate/20200425033724_create_existing_items.rb
```diff
@@ -0,0 +1,8 @@
...
+  def change
+    create_table :existing_items do |t|
+      t.string :idempotent_key
+      t.integer :item_type
```
jtmckibb, line 5: I'd just call this type instead of item_type since we already have an attribute called item_type and I wouldn't wanna confused it

app/models/existing_item.rb
```diff
@@ -0,0 +1,5 @@
...
+  validates_uniqueness_of :idempotent_key
+  enum item_type: [:donation, :giftcard, :refund]
```
jtmckibb, line 4: `charges_create`, and `webhooks_create_refund_updated` is all we need

db/migrate/20200425033724_create_existing_items.rb
jtmckibb: because the question you're asking is: "Has this event happened already?"

jtmckibb: Or maybe `ExistingEvent`

jtmckibb: whatever you think — I think of Transactions as something that you make with the database, and not all calls to our API will lead to a Transaction

app/models/existing_item.rb
```diff
@@ -0,0 +1,5 @@
+class ExistingItem < ApplicationRecord
+  validates_presence_of :idempotent_key, :item_type
+  validates_uniqueness_of :idempotent_key
```
jtmckibb, line 3: You should add a spec that tests that this is unique

jtmckibb: and you also should add a factory

danthaman44:
@jtmckibb Addressed your feedback! All tests pass:
`danthaman-mac:ruby danthaman$ heroku local:run bundle exec rspec
.......**...*.....**.........................................................***.....................................................

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) CustomerMailer add some examples to (or delete) /Users/danthaman/Workspace/ruby/spec/mailers/customer_mailer_spec.rb
     # Not yet implemented
     # ./spec/mailers/customer_mailer_spec.rb:4

  2) DonationDetail add some examples to (or delete) /Users/danthaman/Workspace/ruby/spec/models/donation_detail_spec.rb
     # Not yet implemented
     # ./spec/models/donation_detail_spec.rb:4

  3) GiftCardAmount add some examples to (or delete) /Users/danthaman/Workspace/ruby/spec/models/gift_card_amount_spec.rb
     # Not yet implemented
     # ./spec/models/gift_card_amount_spec.rb:4

  4) Item add some examples to (or delete) /Users/danthaman/Workspace/ruby/spec/models/item_spec.rb
     # Not yet implemented
     # ./spec/models/item_spec.rb:4

  5) Location add some examples to (or delete) /Users/danthaman/Workspace/ruby/spec/models/location_spec.rb
     # Not yet implemented
     # ./spec/models/location_spec.rb:4

  6) Gift Cards API POST /gift_cards
     # No reason given
     # ./spec/requests/gift_cards_spec.rb:10

  7) Gift Cards API GET /gift_cards/:id
     # No reason given
     # ./spec/requests/gift_cards_spec.rb:40

  8) Gift Cards API PUT /gift_cards/:id
     # No reason given
     # ./spec/requests/gift_cards_spec.rb:66


Finished in 15.42 seconds (files took 0.90486 seconds to load)
133 examples, 0 failures, 8 pending`

jtmckibb:
Brilliant 👏

PR #64, opened 2020-04-25
Add vendor/bundle to gitignore

Yong-L:
Does anyone know why the default `bundle install` not install it into vendor/bundle even though there is a config file listed in the project directory?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@
 # Ignore intellij stuff
 .idea/
 ruby.iml
+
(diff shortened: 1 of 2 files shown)


Yong-L:
Ah ignore the above comment about why it wasn't working... It was because bundle config was generated when I ran the command. Why are bundler configurations ignored?

Yong-L:
Ah just realized that `bundle` config is system specific but it is recommended to install project based dependencies per folder in the vendor folder...

PR #66, opened 2020-04-25
Create square_location_id as part of the model definition

Yong-L:
This enables fetching the seller by square_location_id and storing it in
the frontend so that the id can then be referenced when the charge is
made to update square

### DIFF
--- a/app/models/seller.rb
+++ b/app/models/seller.rb
@@ -3,6 +3,9 @@ class Seller < ApplicationRecord
   has_many :locations, dependent: :destroy
   has_many :menu_items, dependent: :destroy
 
+  validates_presence_of :seller_id
+  validates_presence_of :square_location_id
+
   validates_inclusion_of :founded_year, :in => 1800..2020
   validates_uniqueness_of :seller_id
   validates_uniqueness_of :square_location_id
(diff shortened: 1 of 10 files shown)


Yong-L:
Do you have any reference code testing required fields vs non-required?

jtmckibb:
LOL I can't comment on the comments because Stanley accidentally closed the PR but here: https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb

app/models/seller.rb
```diff
@@ -5,6 +5,7 @@ class Seller < ApplicationRecord
 
   validates_inclusion_of :founded_year, :in => 1800..2020
   validates_uniqueness_of :seller_id
+  validates_uniqueness_of :square_location_id
```
jtmckibb, line 8: also wanna validate the presence of too :) 

https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb

Yong-L: Haha no worries

PR #68, opened 2020-04-25
Move seller_id to the charges api body instead of per line_item

Yong-L:
Make the seller_id to the body of the api instead of per line item so that we can use that to reference the square_location_id


Yong-L:
Made the frontend changes: https://github.com/sendchinatownlove/sendchinatownlove.github.io/pull/74

jtmckibb:
Looks great!

PR #70, opened 2020-04-26
Enforce Rubocop

jtmckibb:


.rubocop.yml
```diff
@@ -0,0 +1,120 @@
...
+    - 'vendor/**/*'
+    - 'spec/fixtures/**/*'
+    - 'tmp/**/*'
+    - '.git/**/*'
+    - 'Gemfile'
+    - 'Gemfile.lock'
+    - 'Rakefile'
+    - 'db/schema.rb'
+    - 'bin/**/*'
+  TargetRubyVersion: 2.4
```
Yong-L, line 19: we're using 2.6.x

jtmckibb: good catch

PR #71, opened 2020-04-26
[DO NOT SUBMIT] mailer test repro

juliexxia:
Here's a quick repro for testing the mailer functionality. Copy this branch and run

$ ./post_webhook.sh

The changes in webhook_controller are pretty dirty, I just grabbed and hardcoded in an old transaction's info. It'll fail eventually because it's an old transaction but it should run through the email functionality before then. 

The transaction it's running goes to my email but you should be able to see results in the mailgun dashboard which you can read by running

$ heroku addons:open mailgun

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,7 +46,7 @@ group :development, :test do
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
   gem 'rspec-rails', '~> 3.5'
   gem 'faker'
-  gem 'pry', '~> 0.12.2'
+  gem 'pry-byebug'
 end
 
 group :development do
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -75,7 +75,7 @@ Rails.application.configure do
       :address        => "smtp.mailgun.org",
       :user_name      => "postmaster@sandbox6657e75f5d664a16921d418cd676ac5c.mailgun.org",
       :password       => "69cac0b97a813a000fd8d198c596bfab-f135b0f1-ce7a2435",
-      :domain         => 'sandbox74151aabc4114674a95154526cf10e4f.mailgun.org',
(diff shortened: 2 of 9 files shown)


Gemfile
```diff
@@ -46,7 +46,7 @@ group :development, :test do
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
   gem 'rspec-rails', '~> 3.5'
   gem 'faker'
-  gem 'pry', '~> 0.12.2'
+  gem 'pry-byebug'
```
jtmckibb, line 49: ooo what's this

juliexxia: it fixes a bug I was encountering where I couldn't use the 'next' and 'continue' pry commands

config/environments/production.rb
```diff
@@ -64,9 +64,20 @@
...
   # Set this to true and configure the email server for immediate delivery to raise delivery errors.
-  # config.action_mailer.raise_delivery_errors = false
+  config.action_mailer.raise_delivery_errors = true
+
+  config.action_mailer.smtp_settings = {
+      :port           => 587,
+      :address        => "smtp.mailgun.org",
+      :user_name      => "postmaster@sandbox6657e75f5d664a16921d418cd676ac5c.mailgun.org",
+      :password       => "69cac0b97a813a000fd8d198c596bfab-f135b0f1-ce7a2435",
+      :domain         => 'sendchinatownlove.herokuapp.com,
```
stanleyzhengnyc, line 78: i think this caused our account to be frozen, sorry for unclear instructions

juliexxia: oh boy - sorry! Though I don't think these prod codes were ever actually accessed because I was running through test and never merged this

juliexxia: by test I mean the config/environments/development.rb

PR #72, opened 2020-04-27
Use pry-byebug

juliexxia:
Adds `step`, `next`, `finish` and `continue` commands and breakpoints to Pry using byebug.

To use, invoke pry normally. No need to start your script or app differently.

```
def some_method
  binding.pry          # Execution will stop here.
  puts 'Hello World'   # Run 'step' or 'next' in the console to move here.
end
```

https://www.rubydoc.info/gems/pry-byebug/2.0.0


jtmckibb:
I was wondering why those didn't work

juliexxia:
(:

PR #78, opened 2020-04-29
Refactor out SquareClient to services module

Yong-L:
Per rails best practices, we should be using `services` convention

https://www.toptal.com/ruby-on-rails/rails-service-objects-tutorial

### DIFF
--- a/app/controllers/charges_controller.rb
+++ b/app/controllers/charges_controller.rb
@@ -32,7 +32,7 @@ class ChargesController < ApplicationController
     payment =
       if charge_params[:is_square]
         create_square_payment_request(
-          source_id: charge_params[:nonce],
+          nonce: charge_params[:nonce],
           amount: amount,
           note: description,
           email: email,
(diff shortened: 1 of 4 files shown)


app/controllers/charges_controller.rb
```diff
@@ -107,34 +107,27 @@ def validate(seller_id:, line_items:)
   def generate_description(seller_name:, item_types:)
     description = 'Thank you for supporting ' + seller_name + '.'
     if item_types.include? 'gift_card'
-      description +=
-        ' Your gift card(s) will be emailed to you when the seller opens back up.'
+      description += ' Your gift card(s) will be emailed to you when the seller opens back up.'
```
jtmckibb, line 110: Can you run `rubocop -a` to make sure this line isn't too long?

app/controllers/charges_controller.rb
```diff
@@ -107,34 +107,27 @@ def validate(seller_id:, line_items:)
...
-      source_id: source_id,
-      idempotency_key: SecureRandom.uuid,
-      amount_money: { amount: amount, currency: 'USD' },
-      buyer_email_address: email,
-      note: note,
-      location_id: square_location_id
-    }
-
-    api_response = api_client.payments.create_payment(body: request_body)
+    api_response = SquareManager::PaymentCreator.call(
```
jtmckibb, line 122: Wow this is so much cleaner

app/controllers/charges_controller.rb
Yong-L: Just broke it up. I'm not too sure what the best practice is for ruby for breaking up lines haha

jtmckibb: My own preference is to try and keep the string together so that it's easier to grep — but idc really lol

jtmckibb: usually I just ignore it if it's just a long string

Yong-L: Only problem was that even with the previous change the line was at 82 which exceeded the default 80. Maybe we should change the default limit to 100 or so

PR #80, opened 2020-04-30
Pretty emails local repro

juliexxia:
Testing locally:

0. Pull a copy of this branch
1. Update the file to send to your own email (search for jxiarizard@gmail.com to replace mine)
2. Paste in our heroku API key (be careful not to commit it back!) - ask me, justin or stanley for it
3. Run the server
`$ heroku local web`
4. Send a POST to the webhooks controller
`$ curl --location --request POST 'http://localhost:5000/webhooks'`


app/controllers/webhooks_controller.rb
```diff
@@ -5,17 +5,79 @@
...
+    api_url = "https://api:#{api_key}@api.mailgun.net/v2/m.sendchinatownlove.com/messages"
+
+    amount = 500
+    merchant = "Joolie's Bakery"
+    # receipt_id = "blerghllejs"
+    # receipt_url = "coolmath.com"
+
+    amount_string = '%.2f' % ((amount.to_f)/100)
+
+    html = "<!DOCTYPE html>" +
```
jtmckibb, line 19: Wait this is miserable. Is it possible to write it has an html file and write a script to convert it to this upon deploy? @stanleyzhengnyc If I wrote a Python script or Perl script or something, can we run it as a part of the build?

jtmckibb: Matter of fact, we can just write an HTML file and read it for each email. It's not efficient, but we're talking like 3-6 donations a day.

stanleyzhengnyc: not a rails expert we can load it as a HTML view as an ERB template,we should be able to just load and drop a new file there
like here 
https://github.com/sendchinatownlove/ruby/blob/master/app/views/layouts/receipt.html_A.mjs
that being said its just a rail blocker we're not really familiar with; I looked into it and im not 100% sure how to do it 

https://guides.rubyonrails.org/layouts_and_rendering.html

stanleyzhengnyc: @Jeremy-Xue are you familiar more with rails how to get this email html mailer view

juliexxia: Yeah totally agree this is sad state of affairs. I don't know how to use a parse-from-html helper that can also inserting our own vars into the html (like seller name, etc). 

The thing is, the smtp/Action Mailer method has all this capability built right in - using an html.erb view file that can take in vars. We just still have those auth issues with the smtp method. The receipts are so cute I thought it might be worth getting it out in this format and then coming back to this later and trying again to move it to smtp when other stuff isn't so pressing, maybe trying to get dev help from mailgun or something. 

@Jeremy-Xue if you know an http api way to do with with a mailer view that would be so awesome!

Jeremy-Xue: Taking a look at this rn!

Jeremy-Xue: I've done something similar to this with django's mailer but haven't used a rails mailer or mailgun before :( 

I think I agree with the idea to deploy the current version come back to it later

michello: Maybe this is too late but Ruby has a gem for templates: https://github.com/chamnap/liquid-rails. 

I use it when I styled email templates for work. You can also define `drop` files which are basically where you would declare variables that you can use in your liquid template files.

jacoblb64:

PR #81, opened 2020-05-02
Implemented server-side charge dedup

danthaman44:
1. Rename `type` column in `existing_events` table to `item_type` because `type` is a reserved keyword in ActiveRecord https://stackoverflow.com/questions/17879024/activerecordsubclassnotfound-the-single-table-inheritance-mechanism-failed-to

2. Return an error if there is a charge request with the same `idempotency_key` as a previous charge. This should be rare now that we have client-side deduplication

Test plan:
Verified HTTP response for duplicate charge
Screen Shot 2020-05-01 at 11 41 10 PM

### DIFF
--- a/app/controllers/charges_controller.rb
+++ b/app/controllers/charges_controller.rb
@@ -5,6 +5,17 @@ require 'securerandom'
 class ChargesController < ApplicationController
   # POST /charges
   def create
+    # Validate this not a duplicate charge
+    idempotency_key = charge_params[:idempotency_key]
+    existing_event =
+      ExistingEvent.new(
+        idempotency_key: idempotency_key,
+        event_type: 'charges_create'
+      )
+    unless existing_event.save
+      raise DuplicateChargeError, 'This charge has already been processed'
+    end
+
     line_items = charge_params[:line_items].map(&:to_h)
 
     seller_id = charge_params[:seller_id]
@@ -51,12 +62,14 @@ class ChargesController < ApplicationController
     params.require(:is_square)
     params.require(:nonce) if params[:is_square]
     params.require(:name)
+    params.require(:idempotency_key)
     params.permit(
       :email,
       :nonce,
       :is_square,
       :name,
       :seller_id,
+      :idempotency_key,
       line_items: [%i[amount currency item_type quantity]]
     )
   end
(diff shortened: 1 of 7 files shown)


danthaman44:
This breaks basically all the Charges tests...need to update them

app/controllers/charges_controller.rb
```diff
@@ -5,6 +5,13 @@
 class ChargesController < ApplicationController
   # POST /charges
   def create
+    # Validate this not a duplicate charge
+    idempotency_key = charge_params[:idempotency_key]
+    existing_event = ExistingEvent.new(idempotency_key: idempotency_key, event_type: 'charges_create')
+    unless existing_event.save
+      return json_response(existing_event.errors, :bad_request)
```
jtmckibb, line 12: Could you throw new Exception here instead? And handle it in the ExceptionHandler?

app/controllers/charges_controller.rb
```diff
@@ -5,6 +5,13 @@
 class ChargesController < ApplicationController
   # POST /charges
   def create
+    # Validate this not a duplicate charge
+    idempotency_key = charge_params[:idempotency_key]
+    existing_event = ExistingEvent.new(idempotency_key: idempotency_key, event_type: 'charges_create')
+    unless existing_event.save
```
jtmckibb, line 11: Mind adding a unit test to the charges_spec here?

app/controllers/charges_controller.rb
danthaman44: Good call!

app/controllers/charges_controller.rb
danthaman44: Kinda struggling with...this is the test im envisioning, but it doesn't pass...not sure whats the problem
```
    context 'with a duplicate payment request' do
      let(:line_items) do
        [
          {
            amount: 50,
            currency: 'usd',
            item_type: 'gift_card',
            quantity: 1,
            seller_id: seller_id
          }
        ]
      end

      before { post '/charges', params: params, as: :json }

      it 'records charge events' do
        allow_any_instance_of(ExistingEvent).to receive(:save).and_return(false)
        expect(response).to have_http_status(400)
      end
    end
```

danthaman44: Done

spec/requests/charges_spec.rb
```diff
@@ -172,6 +174,21 @@
...
+        [{ currency: 'usd', item_type: 'gift_card', quantity: 1 }]
+      end
+
+      before do
+        post '/charges', params: params, as: :json
+        allow_any_instance_of(ExistingEvent).to receive(:save).and_return(false)
+      end
+
+      it 'returns status code 422' do
+        expect(response).to have_http_status(422)
```
jtmckibb, line 188: nice! tyyy

PR #82, opened 2020-05-02
Cleanup unused stripe code in controller and specs

Yong-L:
Cleaning up unused stripe code


jtmckibb:
You can delete the keys in the .env file as well. Also if you could delete the migrations via a migration, that would also be awesome

Yong-L:
so just drop the columns?

jtmckibb:
Yeah you can just drop the columns bc we have no Stripe PaymentIntents in prod

jtmckibb:
Actually maybe we should keep them just in case —

Yong-L:
In case of like backwards compat issues?

PR #83, opened 2020-05-02
add merchant_payout model

juliexxia:
No tests yet - please review for model content! 

TODO
- [ ] tests
- [ ] doc fields (in schema.rb?)

### DIFF
--- a/db/migrate/20200502130958_create_merchant_payouts.rb
+++ b/db/migrate/20200502130958_create_merchant_payouts.rb
@@ -1,0 +1,21 @@
+class CreateMerchantPayouts < ActiveRecord::Migration[6.0]
+  def change
+    create_table :merchant_payouts do |t|
+      t.integer :payout_type, default: 0, null:false
+      t.integer :total_amount, null: false
+      t.integer :account_number, null: false
+      t.date :payment_delivered
+
+      # for check payments
+      t.integer :check_number
+      t.date :payment_cashed
+
+      # for cash payments
+      t.date :cash_withdrawn
+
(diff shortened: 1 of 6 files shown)


db/migrate/20200502130958_create_merchant_payouts.rb
```diff
@@ -0,0 +1,21 @@
+class CreateMerchantPayouts < ActiveRecord::Migration[6.0]
+  def change
+    create_table :merchant_payouts do |t|
+      t.integer :payout_type, default: 0, null:false
+      t.integer :total_amount, null: false
+      t.integer :account_number, null: false
+      t.date :payment_delivered
```
jtmckibb, line 7: Could we add a "note" field?

db/migrate/20200502130958_create_merchant_payouts.rb
```diff
@@ -0,0 +1,21 @@
+class CreateMerchantPayouts < ActiveRecord::Migration[6.0]
+  def change
+    create_table :merchant_payouts do |t|
+      t.integer :payout_type, default: 0, null:false
+      t.integer :total_amount, null: false
+      t.integer :account_number, null: false
```
jtmckibb, line 6: Not sure we wanna keep our account numbers in our DB. Could this be a string instead? Right now we have two bank accounts we're paying out from.

Yong-L:
this might make it easier for you to annotate the models: https://github.com/ctran/annotate_models

jacoblb64:

PR #86, opened 2020-05-03
Add Globalize so that we can support Chinese translations

jtmckibb:
Review by commit, because I also fixed a broken payment intent test

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,8 +29,9 @@ gem 'sinatra'
 # Square
 gem 'square.rb'
 
-# Generates hashes
-# gem 'digest'
+# Translations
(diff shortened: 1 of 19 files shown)


Yong-L:
When we generate the translation, what happens if the name and story changes? Will we have to run a command to regenerate it?

jtmckibb:
We have to POST and PUT the story for each locale @Yong-L. Right now, we only have the English locale, so if you GET the Chinese locale it respond with null. In the future, if we change the English, well have to manually PUT the Chinese, which will be translated by Seller Empathy. Seller Empathy are the ones who currently put in merchant stories for the English version, so they'll also just be in charge of also keeping the Chinese version up to date.

PR #87, opened 2020-05-03
annotate models, tests, and config/routes.rb

juliexxia:
Using annotate: https://github.com/ctran/annotate_models

There's apparently a way to auto-update annotations every time `db:migrate` is run but I errored out trying to make that happen so just added instructions to the readme instead. I figure it's okay to manually run intermittently. 

Thanks @Yong-L !

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,62 @@
 # frozen_string_literal: true
 
+# == Route Map
+#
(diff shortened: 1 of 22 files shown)


config/routes.rb
```diff
@@ -1,5 +1,62 @@
 # frozen_string_literal: true
 
+# == Route Map
```
jtmckibb, line 3: OMG this is amazing!  T-T

PR #91, opened 2020-05-06
Add URI parsing to parse out `RAILS_WEBHOOK_URL`

Yong-L:


config/environments/development.rb
```diff
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
 
+require 'uri'
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in
   # config/application.rb.
+  config.hosts << URI.parse(ENV['RAILS_WEBHOOK_URL']).host
```
jtmckibb, line 8: Do you have to include the /webhooks portion of the URL?

Yong-L: We do have to have the /webhooks portion for validation checking or else it fails

We could include the just the host in our env and append https:// and /webhooks in our code. What do you think is the best approach?

PR #93, opened 2020-05-07
working on giftcard viewer

ryannamgung:


app/jobs/sellers_helper.rb
```diff
@@ -11,8 +11,22 @@ def self.generate_seller_json(seller:)
     seller
   end
 
+
+  def self.retrieve_amount_in_card(gift_card_id:)
```
jtmckibb, line 15: This method is already defined if you `git checkout master` `git pull` `git checkout ryannamgung` `git rebase mater`

jtmckibb: In my opinion, you should create a separate GiftCardHelper and add the method there

app/controllers/gift_cards_controller.rb
```diff
@@ -8,21 +8,15 @@ class GiftCardsController < ApplicationController
...
-  end
+    #need to get the gift_card_detail of item and feed its ID into gift card recent
+    SellersHelper.gift_card_recent()
 
-  # PUT /gift_cards/:id
-  def update
-    # only allows updating amounts of gift cards
-    original_amount = @gift_card.amount
-    new_amount = update_params[:amount].to_i
+    current_amount = json_response(item)
```
jtmckibb, line 14: Change this to `json_response(item)`

PR #94, opened 2020-05-07
Add recent_contribution controller

Yong-L:


app/controllers/recent_contribution_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+# Controller to fetch most recent contribution
+class RecentContributionController < ApplicationController
```
jtmckibb, line 4: I think this might should be called ItemsController, and we should add a parameter which allows you to filter by Seller — just to follow general REST standards

jtmckibb: If we go with this route, we'll put ourselves in a position where we'll have a new endpoint for every feature. I'm no expert at API design, but I usually reference Stripe's API as an example when designing APIs

Yong-L: Good point! I’d guess having a generic items endpoint which takes in the query you’d like would be most optimal? Such as /sellers/items?sort=desc&limit=1 ?

jtmckibb: Yeah totally

jtmckibb:
Looks like you might have accidentally pulled other commits in while rebasing

db/migrate/20200510164927_add_amount_to_item.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Migration to add amount of the item that was initially at payment_intent item
+class AddAmountToItem < ActiveRecord::Migration[6.0]
```
jtmckibb, line 4: Why are we adding amount to Item? Amount is variable, and is derived from the GiftCardAmount on the DonationDetail

db/schema.rb
```diff
@@ -56,6 +56,7 @@
     t.datetime "updated_at", precision: 6, null: false
     t.bigint "payment_intent_id", null: false
     t.boolean "refunded", default: false
+    t.decimal "amount", null: false
```
jtmckibb, line 59: I don't think we want to do this since Amount is variable for GiftCards. We don't save the amount directly on the Item so that we can save a history. See here: https://docs.google.com/drawings/d/13ln9-dCYl_wUxLVkMfCH5VFOe9BgkzY0Qu8V-TTl0Fs/edit

Yong-L:
> Looks like you might have accidentally pulled other commits in while rebasing

Ah how do I fix that

db/schema.rb
Yong-L: I always thought that GiftCardAmount was going to be used to track the balance of the gift card

Yong-L:
NVM I figured it out :)

db/migrate/20200510164927_add_amount_to_item.rb
Yong-L: I thought that since GiftCardAmount would be used to track the balance, once we enable the gift card redemption and we start decrementing, it might get weird

app/controllers/webhooks_controller.rb
```diff
@@ -227,6 +227,10 @@ def generate_seller_gift_card_id(seller_id:)
     raise CannotGenerateUniqueHash, 'Error generating unique gift_card_id'
   end
 
+  def generate_seller_gift_card_id_hash
```
jtmckibb, line 230: This also doesn't look it like should be a part of this PR hmmm

app/controllers/items_controller.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class ItemsController < ApplicationController
+
+  before_action :set_seller
+
+  def index
+    query = Validate::ItemQuery.new(params)
```
jtmckibb, line 8: This works, but ActionController::Parameters has a .permit and .require method that you can use to validate

Yong-L: Yup was planning to use it but I wanted a wrapper to enable me to use `ActiveModel::Validations`

app/controllers/webhooks_controller.rb
Yong-L: ahhhhhhhhhhhhhh

PR #100, opened 2020-05-10
Added user model + db migration to power email subscriptions

danthaman44:
```
danthaman-mac:ruby danthaman$ heroku local:run bundle exec rspec spec/models/user_spec.rb
...

Finished in 0.17486 seconds (files took 1.67 seconds to load)
3 examples, 0 failures
```

### DIFF
--- a/db/migrate/20200510012452_create_users.rb
+++ b/db/migrate/20200510012452_create_users.rb
@@ -1,0 +1,8 @@
+class CreateUsers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :users do |t|
+      t.string :email, optional: true
+      t.boolean :is_subscribed
+    end
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,0 +1,12 @@
+# == Schema Information
+#
+# Table name: users
+#
+#  id            :bigint           not null, primary key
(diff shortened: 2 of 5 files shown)


db/migrate/20200510012452_create_users.rb
```diff
@@ -0,0 +1,8 @@
+class CreateUsers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :users do |t|
+      t.string :email, unique: true
```
danthaman44, line 4: is `unique: true` necessary if the model already enforces uniqueness for the field?

jtmckibb: It's good to have both DB and model enforcement

jtmckibb: But crazy, can we make this optional? I think we designed the same model: See "Contact"

https://docs.google.com/document/d/1pbP1NX2bdH_aZGbBJxwxWUfvIf8npKSn0yrDftBGvz8/edit#

Yong-L:
Nope doesn’t matter for me! LGTM

db/migrate/20200510012452_create_users.rb
danthaman44: Done

PR #103, opened 2020-05-11
Yong/user migration

Yong-L:
Using this PR to automatically run tests for me

### DIFF
--- a/app/controllers/charges_controller.rb
+++ b/app/controllers/charges_controller.rb
@@ -140,6 +140,11 @@ class ChargesController < ApplicationController
     # TODO (yong): Also query to find out the recipient
     contact = Contact.find_or_create_by(email: email)
 
+    if contact.name != name
+      contact.name = name
+      contact.save!
+    end
+
     # Creates a pending PaymentIntent. See webhooks_controller to see what
     # happens when the PaymentIntent is successful.
     PaymentIntent.create!(
--- a/app/controllers/charges_controller.rb
+++ b/app/controllers/charges_controller.rb
@@ -137,15 +137,18 @@ class ChargesController < ApplicationController
     payment = api_response.data.payment
     receipt_url = payment[:receipt_url]
 
+    # TODO (yong): Also query to find out the recipient
+    contact = Contact.find_or_create_by(email: email)
+
     # Creates a pending PaymentIntent. See webhooks_controller to see what
     # happens when the PaymentIntent is successful.
     PaymentIntent.create!(
       square_location_id: square_location_id,
       square_payment_id: payment[:id],
-      email: email,
       line_items: line_items.to_json,
       receipt_url: receipt_url,
-      name: name
+      purchaser: contact,
+      recipient: contact
     )
 
     api_response
--- a/db/migrate/20200511014428_add_email_index_on_user.rb
+++ b/db/migrate/20200511014428_add_email_index_on_user.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddEmailIndexOnUser < ActiveRecord::Migration[6.0]
+  def change
+    # Adding index as we'll be using this to do extensive queries
+    add_index :users, :email
+  end
+end
--- a/db/migrate/20200511200219_add_contact_to_payment_intent_and_item.rb
+++ b/db/migrate/20200511200219_add_contact_to_payment_intent_and_item.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+class AddContactToPaymentIntentAndItem < ActiveRecord::Migration[6.0]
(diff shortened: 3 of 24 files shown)


jtmckibb:
In the future we should update is_subscribed to be an `EmailSubscriber` table which has a bunch of columns like subscribed:Contact, receipt_sent:boolean, gift_card_sent:boolean... or something like that @danthaman44 

We don't have to do this now — we can think about it a bit but might want to migrate the subscription off of the Contact

Yong-L:
Should be good to review now...

db/schema.rb
```diff
@@ -95,6 +112,10 @@
     t.string "email_text"
     t.string "receipt_url"
     t.string "name"
+    t.bigint "purchaser_id"
```
jtmckibb, line 115: This should probably be nonnullable

db/schema.rb
```diff
@@ -95,6 +112,10 @@
...
+    t.bigint "recipient_id"
```
jtmckibb, line 116: same

db/schema.rb
```diff
@@ -56,7 +71,9 @@
     t.datetime "updated_at", precision: 6, null: false
     t.bigint "payment_intent_id", null: false
     t.boolean "refunded", default: false
+    t.bigint "purchaser_id"
```
jtmckibb, line 74: This doesn't have to be nonnullable bc maybe we'll one day combine purchased gift cards into one for organizations to bulk order

db/schema.rb
```diff
@@ -45,7 +58,9 @@
     t.datetime "updated_at", precision: 6, null: false
     t.bigint "item_id", null: false
     t.string "seller_gift_card_id"
+    t.bigint "recipient_id"
```
jtmckibb, line 61: This should be nonnullable

db/migrate/20200511200219_add_contact_to_payment_intent_and_item.rb
```diff
@@ -0,0 +1,42 @@
...
+      payment_intent.save!
+
+      # Item will at most be one so this will run only once
+      payment_intent.items.each do |item|
+        item.purchaser = contact
+        item.save!
+
+        gift_card_detail = item.gift_card_detail
+
+        if gift_card_detail != nil
```
jtmckibb, line 35: you could also do itemType == 1

db/migrate/20200511014428_add_email_index_on_user.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddEmailIndexOnUser < ActiveRecord::Migration[6.0]
+  def change
+    # Adding index as we'll be using this to do extensive queries
```
jtmckibb, line 5: sick

db/schema.rb
Yong-L: rails default `null` to true

https://api.rubyonrails.org/v6.0.3/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_reference

db/schema.rb
Yong-L: same above

db/schema.rb
Yong-L: cool, I'll add in that to the migration and make the model change as well

app/controllers/charges_controller.rb
```diff
@@ -137,15 +137,23 @@ def create_square_payment_request(
     payment = api_response.data.payment
     receipt_url = payment[:receipt_url]
 
+    # TODO (yong): Also query to find out the recipient
+    contact = Contact.find_or_create_by(email: email)
```
danthaman44, line 141: What are your thoughts on putting this in a Service? maybe called ContactRegistration?

danthaman44: I can do that another PR

Yong-L: Sounds good should be something like ContactCreator to follow the or convention I wrongly started

BTW I saw in the other PR that people ask for email subscriptions so I’d imagine that we’d want the default to be false?

danthaman44: hmmm thats a good question, I'm not sure whats the best way to model this. One idea is that if people don't subscribe, we just don't save any information for them at all in the DB, eliminating the need for a `is_subscribed` column.

Yong-L: Good point but since we’d still be leveraging the tbl for donation information when we do create items, we’d still need it

I guess we can keep it to default false for now

jtmckibb:
I’d rather default to true and build them a way to opt out 😅

Yong-L:
Sounds good

PR #105, opened 2020-05-13
readme local webhook updates

juliexxia:
I ~think~ the two steps I took out were unnecessary. FWIW I couldn't start up heroku local web until I did all the other steps so I figure it must not be that important but let me know if I'm missing something!


juliexxia:
Thanks for walking me through it!

PR #106, opened 2020-05-13
Added Contact Registration service

danthaman44:
Service to handle registration of contacts when they make a donation/purchase a voucher
Tested with data sent from front-end https://github.com/sendchinatownlove/sendchinatownlove.github.io/pull/111


danthaman44:
Ughh need to fix the tests

jtmckibb:
A unit test for ContactRegistrator would be great but I won't block it

danthaman44:
I'll try to write a test as a followup!

PR #107, opened 2020-05-14
Add seller <> contact relationship

Yong-L:
This is to enable us to query the `sellers` table when we wish to have a different recipient other than the person who made the contribution.

### DIFF
--- a/app/helpers/sellers_helper.rb
+++ b/app/helpers/sellers_helper.rb
@@ -3,9 +3,9 @@
 module SellersHelper
   def self.generate_seller_json(seller:)
     locations = seller.locations
-    recipient = seller.distributor
+    distributor = seller.distributor
     seller = seller.as_json
-    seller['distributor'] = distributor.as_json unless recipient == nil
+    seller['distributor'] = distributor.as_json unless distributor == nil
     seller['locations'] = locations.as_json
     seller['gift_card_amount'] = calculate_gift_card_amount(
       seller_id: seller['id']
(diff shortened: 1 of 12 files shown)


app/helpers/sellers_helper.rb
```diff
@@ -3,7 +3,9 @@
 module SellersHelper
   def self.generate_seller_json(seller:)
     locations = seller.locations
+    recipient = seller.recipient
     seller = seller.as_json
+    seller['recipient'] = recipient.as_json if recipient != nil
```
jtmckibb, line 8: nit unless recipient.nil?

jtmckibb: Can we call it can_buy_a_meal_for? Right now it's not really clear why a Contact is related to a seller

jtmckibb: or maybe partner? distributor? buy_a_meal_for_partner? buy_a_meal_for_distributor?

app/models/seller.rb
```diff
@@ -40,6 +40,8 @@ class Seller < ApplicationRecord
   has_many :menu_items, dependent: :destroy
   has_many :items, dependent: :destroy
 
+  has_one :recipient, class_name: 'Contact'
```
jtmckibb, line 43: This could either use a better name or more documentation around what it's used for

maybe partner? distributor? buy_a_meal_for_partner? buy_a_meal_for_distributor?

Yong-L: Done! Named it to distributor

Yong-L:
Removing this pull request as github actions seems like it's not working as expected

PR #108, opened 2020-05-14
Add pool donations to webhooks controller

juliexxia:
- [x] write tests
- [x] don't allocate to merchants who don't accept donations

### DIFF
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -2,6 +2,9 @@
 
 require 'rest-client'
 
+# TODO(juliexxia):  document
+POOL_DONATION_SELLER_ID = 'send-chinatown-love'
+
 # TODO(jmckibben): This class needs a lot of refactoring
 class WebhooksController < ApplicationController
   # POST /webhooks
@@ -120,55 +123,87 @@ class WebhooksController < ApplicationController
 
       amount = item['amount']
       seller_id = item['seller_id']
-      merchant_name = Seller.find_by(seller_id: seller_id).name
-      case item['item_type']
-      when 'donation'
-        item = create_item(
-          item_type: :donation,
-          seller_id: seller_id,
-          purchaser: purchaser,
-          payment_intent: payment_intent
-        )
-        create_donation(item: item, amount: amount)
-        begin
-          send_donation_receipt(
-            payment_intent: payment_intent,
-            amount: amount,
-            merchant: merchant_name
-          )
-        rescue StandardError
-        end
-      when 'gift_card'
-        purchaser = payment_intent.purchaser
-        item = create_item(
-          item_type: :gift_card,
-          seller_id: seller_id,
-          purchaser: purchaser,
-          payment_intent: payment_intent
-        )
 
-        recipient = payment_intent.recipient
-        gift_card_detail = create_gift_card(
-          item: item,
-          amount: amount,
-          seller_id: seller_id,
-          recipient: recipient
-        )
+
+      if seller_id.eql?(POOL_DONATION_SELLER_ID)
+        unless item['item_type'].eql?('donation')
+          raise InvalidPoolDonationError,
+                "pool contribution must but be of type 'donation' but found type #{item['item_type']}."
+        end
+
+        # calculate amount per merchant
+        # This will break if we ever have zero merchants but are still
+        # accepting pool donations.
+        amount_per = (amount.to_f/(Seller.count - 1).to_f).round
+
+        Seller.all.each do |seller|
+          unless seller.seller_id.eql?(POOL_DONATION_SELLER_ID)
+            new_item = create_item(
+                item_type: :donation,
+                seller_id: seller.seller_id,
+                purchaser: purchaser,
+                payment_intent: payment_intent
+            )
+            create_donation(item: new_item, amount: amount_per)
+          end
+        end
         begin
-          send_gift_card_receipt(
-            payment_intent: payment_intent,
-            amount: amount,
-            merchant: merchant_name,
-            receipt_id: gift_card_detail.seller_gift_card_id
+          send_pool_donation_receipt(
+              payment_intent: payment_intent,
+              amount: amount,
           )
         rescue StandardError
         end
       else
-        raise(
-          InvalidLineItem,
-          'Unsupported ItemType. Please verify the line_item.name.'
-        )
+        merchant_name = Seller.find_by(seller_id: seller_id).name
+        case item['item_type']
+        when 'donation'
+          item = create_item(
+              item_type: :donation,
+              seller_id: seller_id,
+              purchaser: purchaser,
+              payment_intent: payment_intent
+          )
+          create_donation(item: item, amount: amount)
+          begin
+            send_donation_receipt(
+                payment_intent: payment_intent,
+                amount: amount,
+                merchant: merchant_name
+            )
+          rescue StandardError
+          end
+        when 'gift_card'
+          item = create_item(
+              item_type: :gift_card,
+              seller_id: seller_id,
+              purchaser: purchaser,
+              payment_intent: payment_intent
+          )
+
+          gift_card_detail = create_gift_card(
+              item: item,
+              amount: amount,
+              seller_id: seller_id,
+              recipient: recipient
+          )
+          begin
+            send_gift_card_receipt(
+                payment_intent: payment_intent,
+                amount: amount,
+                merchant: merchant_name,
+                receipt_id: gift_card_detail.seller_gift_card_id
+            )
+          rescue StandardError
+          end
+        else
+          raise(
+              InvalidLineItem,
+              'Unsupported ItemType. Please verify the line_item.name.'
+          )
+        end
       end
+
     end
 
     # Mark the payment as successful once we've recorded each object purchased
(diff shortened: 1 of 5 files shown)


app/controllers/webhooks_controller.rb
```diff
@@ -120,55 +123,87 @@ def handle_payment_intent_succeeded(
...
+              amount: amount,
           )
         rescue StandardError
         end
       else
-        raise(
-          InvalidLineItem,
-          'Unsupported ItemType. Please verify the line_item.name.'
-        )
+        merchant_name = Seller.find_by(seller_id: seller_id).name
```
juliexxia, line 158: This whole bit in the else block is essentially the code from before, git diff just getting confused

app/controllers/webhooks_controller.rb
```diff
@@ -2,6 +2,9 @@
 
 require 'rest-client'
 
+# TODO(juliexxia):  document
+POOL_DONATION_SELLER_ID = 'send-chinatown-love'
```
jtmckibb, line 6: Can we move this into a constant in the Sellers model?

jtmckibb: Also could you add this special Seller into our seeds.rb?

app/controllers/webhooks_controller.rb
```diff
@@ -120,55 +123,87 @@ def handle_payment_intent_succeeded(
...
-        )
-        create_donation(item: item, amount: amount)
-        begin
-          send_donation_receipt(
-            payment_intent: payment_intent,
-            amount: amount,
-            merchant: merchant_name
-          )
-        rescue StandardError
+
```
jtmckibb, line 126: nit extra newline

app/controllers/webhooks_controller.rb
```diff
@@ -120,55 +123,87 @@ def handle_payment_intent_succeeded(
...
 
-        recipient = payment_intent.recipient
-        gift_card_detail = create_gift_card(
-          item: item,
-          amount: amount,
-          seller_id: seller_id,
-          recipient: recipient
-        )
+        # calculate amount per merchant
+        # This will break if we ever have zero merchants but are still
```
jtmckibb, line 135: lol nice

app/controllers/webhooks_controller.rb
```diff
@@ -120,55 +123,87 @@ def handle_payment_intent_succeeded(
...
+        # accepting pool donations.
+        amount_per = (amount.to_f/(Seller.count - 1).to_f).round
```
jtmckibb, line 137: Can you pull this into a separate method? It needs a bit more logic built in. It should round to the nearest cent since we can't have fractions of cents, then give the extra cents to each merchant in a random order (sort of like Venmo does when you use their divide calculator)

app/controllers/webhooks_controller.rb
```diff
@@ -120,55 +123,87 @@ def handle_payment_intent_succeeded(
...
+            create_donation(item: new_item, amount: amount_per)
+          end
+        end
         begin
-          send_gift_card_receipt(
-            payment_intent: payment_intent,
-            amount: amount,
-            merchant: merchant_name,
-            receipt_id: gift_card_detail.seller_gift_card_id
+          send_pool_donation_receipt(
```
jtmckibb, line 151: lol nice I guess this is better than sending a receipt per seller

app/controllers/webhooks_controller.rb
```diff
@@ -120,55 +123,87 @@ def handle_payment_intent_succeeded(
...
-          recipient: recipient
-        )
+        # calculate amount per merchant
+        # This will break if we ever have zero merchants but are still
+        # accepting pool donations.
+        amount_per = (amount.to_f/(Seller.count - 1).to_f).round
+
+        Seller.all.each do |seller|
+          unless seller.seller_id.eql?(POOL_DONATION_SELLER_ID)
+            new_item = create_item(
```
jtmckibb, line 141: Line 141 to 147 could be abstracted out and shared with the other donation code-path

app/controllers/webhooks_controller.rb
```diff
@@ -235,6 +270,27 @@ def generate_seller_gift_card_id_hash
...
+           '<p> You\'re donation will be distributed evenly between our merchants. Sending ' \
+           '  thanks from us and from Chinatown for your support! </p>' \
+           '<p> Love,<p>' \
+           '<p> the Send Chinatown Love team</p>' \
+           '</body>' \
+           '</html>'
+    send_receipt(to: payment_intent.email, html: html)
+  end
+
   # rubocop:disable Layout/LineLength
```
jtmckibb, line 294: nit: Delete this line since it's already disabled above

app/controllers/webhooks_controller.rb
```diff
@@ -120,55 +123,87 @@ def handle_payment_intent_succeeded(
...
-          send_gift_card_receipt(
-            payment_intent: payment_intent,
-            amount: amount,
-            merchant: merchant_name,
-            receipt_id: gift_card_detail.seller_gift_card_id
+          send_pool_donation_receipt(
+              payment_intent: payment_intent,
+              amount: amount,
           )
         rescue StandardError
```
jtmckibb, line 155: Mind adding a comment of why this is here? Otherwise it looks like a random rescue. Ideally the receipt sending will be on its own thread (not the main thread)

app/controllers/webhooks_controller.rb
```diff
@@ -120,55 +123,87 @@ def handle_payment_intent_succeeded(
...
-            amount: amount,
-            merchant: merchant_name
-          )
-        rescue StandardError
+
+
+      if seller_id.eql?(POOL_DONATION_SELLER_ID)
+        unless item['item_type'].eql?('donation')
+          raise InvalidPoolDonationError,
+                "pool contribution must but be of type 'donation' but found type #{item['item_type']}."
```
Yong-L, line 131: this could probably be in the `validate` module

app/controllers/webhooks_controller.rb
Jeremy-Xue: I made a seller in seeds.rb that I can send over! Might need help with making a new square location for it though (a few of the fields are placeholders such as num_employees, business type, image url <- I left this as shunfa since it won't actually be user facing but we can put another dummy)

`
  {
    seller_id: 'send-chinatown-love',
    cuisine_name: 'Donation Pool',
    name: 'Send Chinatown Love Donation Pool',
    story: "If you can't decide which merchant to donate to, the Donation Pool will split your donation evenly amongst all of our available merchants",
    accept_donations: true,
    sell_gift_cards: true,
    owner_name: 'Send Chinatown Love',
    owner_image_url: 'assets/shunfa-bakery-logo.png',
    target_amount: 1_000_000,
    summary: nil,
    hero_image_url: 'assets/shunfa-bakery-hero.png',
    progress_bar_color: nil,
    business_type: 'Family-owned',
    num_employees: 42,
    founded_year: 2020,
    website_url: nil,
    menu_url: nil,
    square_location_id: 'E4R1NCMHG7B2Y'
  }
`

app/controllers/webhooks_controller.rb
```diff
@@ -235,6 +270,27 @@ def generate_seller_gift_card_id_hash
     ('a'..'z').to_a.sample(5).join
   end
 
+  # rubocop:disable Layout/LineLength
+  def send_pool_donation_receipt(payment_intent:, amount:)
+    amount_string = format('%.2f', (amount.to_f / 100))
```
jtmckibb, line 275: mind running `rubocop -a` and resolving?

spec/requests/webhooks_spec.rb
```diff
@@ -24,7 +24,51 @@
         line_items: line_items
       )
     end
-    let(:seller) { create :seller }
+    let!(:seller_1) do
```
jtmckibb, line 27: nit: seller1

spec/requests/webhooks_spec.rb
```diff
@@ -24,7 +24,51 @@
...
+      Seller.create(
```
jtmckibb, line 28: You should use the Seller factory `create :seller, seller_id: 'shunfa-bakery', ...` but you probably don't need to add the rest of the stuff since it'll be randomized

spec/requests/webhooks_spec.rb
```diff
@@ -24,7 +24,51 @@
...
+          business_type: 'small-biz',
+          num_employees: 5,
+          founded_year: 1850,
+          website_url: 'https://www.youtube.com/watch?v=jIIuzB11dsA',
+          menu_url: 'https://www.youtube.com/watch?v=jIIuzB11dsA',
+          square_location_id: '1234-abcd'
+      )
+    end
+    let!(:seller_2) do
+      Seller.create(
```
jtmckibb, line 46: Same factory; only add what you need

spec/requests/webhooks_spec.rb
```diff
@@ -24,7 +24,51 @@
...
+          business_type: 'medium-biz',
+          num_employees: 10,
+          founded_year: 1950,
+          website_url: 'https://www.youtube.com/watch?v=C_oACPWGvM4',
+          menu_url: 'https://www.youtube.com/watch?v=C_oACPWGvM4',
+          square_location_id: '4567-efgh'
+      )
+    end
+    let!(:seller_pool) do
+      Seller.create(
```
jtmckibb, line 64: Could you add this into a separate test seeds.rb?

app/controllers/webhooks_controller.rb
juliexxia: Do you mind if I come back and do it after? It's going to auto-fix a bunch of stuff that's not related to this PR and I don't want the logic to get lost in the mix

juliexxia: I fixed all the stuff in the relevant files!

app/controllers/webhooks_controller.rb
jtmckibb: I feel like we should else it'll be an accounting nightmare when cents don't add up and we have to reconcile which splits got rounded up vs which got rounded down. Sorry I know it's a pain! Slash a lot of engineering effort for something so dumb LOL

spec/requests/webhooks_spec.rb
juliexxia: Gotcha yay!

app/controllers/webhooks_controller.rb
jtmckibb: Until the MailChimp goes down and we find out until we end up with a bunch of unpaid-out money in our bank account — or someone messages us wondering when they'll receive their gift card for a merchant that opened up 3 months ago XD

app/controllers/webhooks_controller.rb
jtmckibb: sgtm!

spec/requests/webhooks_spec.rb
juliexxia: Is that something we can hook up to actually seed all tests or just a file people can copy out of manually to seed their own tests?

jtmckibb: It'll seed all tests

jtmckibb: We'll need it here because we should always expect this "special seller" to exist on all environments

juliexxia:
Okay updated to not factor in sellers who don't take donations (also nice because it removes the need for the -1 since the special seller can also just not take donations (:

PR #110, opened 2020-05-15
add send-chinatown-love seller to seeds.rb

juliexxia:
- [ ] get square location id

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -30,7 +30,7 @@
       accept_donations: false,
       sell_gift_cards: false,
       founded_year: 2020,
-      square_location_id: "" # TODO
+      square_location_id: ENV['POOL_SQUARE_LOCATION_ID']
   }
 ].each do |attributes|
   Seller.find_or_initialize_by(seller_id: attributes[:seller_id]).update!(attributes)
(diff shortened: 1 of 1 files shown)


juliexxia:
@Jeremy-Xue have you had a chance to rustle up a square_location_id?

Jeremy-Xue:
> @Jeremy-Xue have you had a chance to rustle up a square_location_id?

1GDT5F6VR86GP

Yup! So I generated this one we can use to test it out (I didn't know what address to put so I put mine) but I think if we want to use it in prod we should probably choose another address.

db/seeds.rb
```diff
@@ -24,6 +24,13 @@
     website_url: nil,
     menu_url: nil,
     square_location_id: 'E4R1NCMHG7B2Y'
+  },
+  {
+      seller_id: 'send-chinatown-love',
+      accept_donations: false,
+      sell_gift_cards: false,
+      founded_year: 2020,
+      square_location_id: "" # TODO
```
jtmckibb, line 33: This should probably be an ENV variable

juliexxia: Should I also make the one for shunfa an ENV var?

PR #114, opened 2020-05-15
Add `is_distribution` flag to charges request

Yong-L:
Added `is_distribution` flag to charges api to determine if we should query `Seller` to fetch the `distributor` and assign that as the `recipient` instead of the `purchaser`.

### DIFF
--- a/app/controllers/charges_controller.rb
+++ b/app/controllers/charges_controller.rb
@@ -32,13 +32,18 @@ class ChargesController < ApplicationController
         sum + item['amount'] * item['quantity']
       end
 
+    is_distribution = !charge_params[:is_distribution].nil? ?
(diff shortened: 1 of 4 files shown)


app/controllers/charges_controller.rb
```diff
@@ -32,13 +32,18 @@ def create
         sum + item['amount'] * item['quantity']
       end
 
+    is_distribution = !charge_params[:is_distribution].nil? ?
```
jtmckibb, line 35: .present?

jtmckibb: `is_distribution = charge_params[:is_distribution] || false`?

Yong-L: ah that's so much cleaner smart

PR #123, opened 2020-05-20
add logging to debug gift card issue

juliexxia:
Can't reproduce this issue locally so adding some logging to try to figure it out in prod.
This will muddy up the heroku logs but I'll remove the non-failure logging when i figure out what's going on

### DIFF
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -191,7 +191,16 @@ class WebhooksController < ApplicationController
               receipt_id: gift_card_detail.seller_gift_card_id
             )
           rescue StandardError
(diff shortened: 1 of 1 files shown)


app/controllers/webhooks_controller.rb
```diff
@@ -191,7 +191,16 @@ def handle_payment_intent_succeeded(
               receipt_id: gift_card_detail.seller_gift_card_id
             )
           rescue StandardError
-            # don't let a failed email bring down the whole POST
+            puts "Gift card email errored out. Printing vars:"
```
jtmckibb, line 194: Rails.logger.info "Gift card email errored out. payment_intent: #{payment_intent} ..."

PR #124, opened 2020-05-20
fix gift card logging

juliexxia:


jtmckibb:
idk why that last one didn't work

juliexxia:
maybe heroku logs doesn't like \n

PR #126, opened 2020-05-20
Distribute the remaining cents between merchants

jtmckibb:
When calculating how much to pay each merchant,
we can't divide cents. We were previously rounding
up. However, that will cause us to have accounting
errors when trying to do things like "figure out how
much fees we paid Square" or when we want to know
how much money we comped for fees to the merchant.

Therefore we added some logic to distribute the
remaining cents.

### DIFF
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -130,7 +130,8 @@ class WebhooksController < ApplicationController
         # calculate amount per merchant
         # This will break if we ever have zero merchants but are still
         # accepting pool donations.
-        amount_per = (amount.to_f / @donation_sellers.count.to_f).ceil
+        amount_per = (amount.to_f / @donation_sellers.count.to_f).floor
+        remainder = amount % @donation_sellers.count
 
         @donation_sellers.each do |seller|
           next if seller.seller_id.eql?(Seller::POOL_DONATION_SELLER_ID)
@@ -139,8 +140,9 @@ class WebhooksController < ApplicationController
             seller_id: seller.seller_id,
             purchaser: purchaser,
             payment_intent: payment_intent,
-            amount: amount_per
+            amount: amount_per + (remainder > 0 ? 1 : 0)
           )
+          remainder -= 1
         end
         begin
           send_pool_donation_receipt(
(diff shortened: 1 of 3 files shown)


app/models/seller.rb
```diff
@@ -35,6 +35,7 @@ class Seller < ApplicationRecord
   # The `seller_id` of the special seller we use to collect pool donations.
   POOL_DONATION_SELLER_ID = 'send-chinatown-love'
 
+  # NOTE: POOL_DONATION_SELLER_ID should not accept donations
```
juliexxia, line 38: This already exists - is this just a note to acknowledge that? maybe s/should/does?

app/controllers/webhooks_controller.rb
```diff
@@ -139,8 +140,9 @@ def handle_payment_intent_succeeded(
             seller_id: seller.seller_id,
             purchaser: purchaser,
             payment_intent: payment_intent,
-            amount: amount_per
+            amount: amount_per + (remainder > 0 ? 1 : 0)
```
juliexxia, line 143: Do you know if @donation_sellers.each is deterministically ordered? If so, won't the remainder always go to the same businesses? I don't think that's a huge problem, but getting the cents to go to random sellers was the bit of complexity I was avoiding implementing in the first place.

jtmckibb: it's not a huge problem, and i don't think matters too much. i could shuffle the sellers, but that would be a nightmare to test, and i don't think they'd fret over the extra cent. If we get 300 pooled donations, it would be a difference of only three dollars

we can merge as is, and maybe we can follow up shuffling the merchants and writing a test for that shuffle

app/models/seller.rb
jtmckibb: yeah this is because as i was reading through the code it wasn't apparent to me, so wanted to make it clear here

app/controllers/webhooks_controller.rb
juliexxia: sgtm

app/models/seller.rb
juliexxia: That makes sense to me - can we rephrase as "does not accept donations"

jacoblb64:
Glad to see this was merged. Not sure if we're dealing with negative numbers at all, but I know that [commercial rounding (away from zero)](https://en.wikipedia.org/wiki/Rounding#Round_half_away_from_zero) is used in a lot of applications to avoid remainder cents.

PR #127, opened 2020-05-22
Debugging PR (DO NOT SUBMIT?)

juliexxia:
I rain `rails db:migrate` to try to fix some console discrepancies between local and heroku that I was having (which worked!) but gave me all these changes - any idea why? and what to do?


Yong-L:
Which changes are you referring to? Like the doc changes?

PR #128, opened 2020-05-22
Juliexxia/refactor emails

juliexxia:
I read some stuff about how pulling helper methods out to a concern is an anti-pattern if that concern is only used for a single controller. That makes sense to me but I sorta think it's worth it to make webhooks_controller less massive and hey maybe we'll use Emails somewhere else eventually. 

Re: testing - didn't add tests because wasn't totally sure what we would test...can you mock methods and make sure the mockers were called in Rspec? Read some articles but our email methods don't really affect any objects - https://glennstovall.com/how-to-test-corns-in-rails-4/


juliexxia:
Woop, fixing some errors one sec

Yong-L:
You could mock out modules and classes using `double`.

Not a rails expert but I did start off using `services` pattern and refered it in my recent PR, might want to consider that as an option. 

I'm not too sure what `conerns` are tbh haha

https://github.com/sendchinatownlove/ruby/pull/129

app/controllers/concerns/emails.rb
```diff
@@ -0,0 +1,114 @@
+# frozen_string_literal: true
+
+module Emails
```
Yong-L, line 3: Highlighted in the comment, maybe this should belong in `services` and we can have a module called `EmailManager`. 

The `EmailManager` module could have a different `Sender` classes that sends out emails differently according to the templatet you've defined below and could call the `EmailManager::Sender` to send out emails so that it'll be easier to digest and unit test.

This could be unit tested using double as well to assert that `EmailSender::Sender.call` was called properly

jtmckibb: Would be good to agree on a single pattern, and it looks like Yong started the trend and it works

juliexxia: Are you imagining we mock out EmailSender::*Sender.call and call webhooks POST to unit test that it's getting called correctly? That doesn't feel super unit test-y to me but it's probably a good test to have and I can't think of anything better off the top of my head haha

juliexxia:
Cool sgtm, factored out to a service instead.

FWIW: https://medium.com/boosterstage/refactoring-code-smells-with-service-objects-88775a7cbaca and search for "concern". This is basically the best resource I found comparing the two models and it's not great haha. Happy to go with services and switch later if things seem to be getting really wonky.

juliexxia:
More interesting reading: https://stackoverflow.com/questions/46457227/concerns-vs-services

app/controllers/concerns/emails.rb
Yong-L: ah for the unit test I was going with the lines of having one sender class and the other specific sender classes such as `gift_card` etc to use that class. But within the scope of `gift_card_sender` you don't have to worry about the logic that actually sends out the email, that could be tested within the sender class. the `gift_card_sender` could just assert that the mocked class was called in order to abstract out the underlying sending logic since it looks like that's the same for all the underlying classes.

Yong-L: Example

```ruby
module EmailManager
  class Sender
    def call
      # do woek
    end
  end
end

module EmailManager
  class GiftCardSender
    def call
      # All the templating stuffo
      EmailManager::Sender.call(template + params)
    end
  end
end

describe EmailManager::GiftCardSender, type: :stuff do

  let(:sender) { double(EmailManager::GiftCardSender) }

  it 'should send out gift card email' do
    EmailManager::GiftCardSender.call(params)

    expect(sender).to have been called
  end
end
```

juliexxia: Ah, so you think we should have a generic Sender class that the specific classes i.e. GiftCardSender instantiate and call methods on? Why do that instead of just having GiftCardSender, DonationSender, PoolDonationSender as classes and just calling methods that belong to the EmailManager module? That's how I have it set up now, for reference. Genuine question I feel like I know nothing about rails best practices haha.

Yong-L: I'm not too sure if it's rails best practice but I thought of it as more abstraction since we could abstract out mocking out `RestClient.post` in the unit test

i.e. not doing assertions like below in the other specs

```ruby
let(:rest_client) { double(RestClient)

expect(RestClient.post).to have_been_called_with(params)
```

I think this is why generally it might be better to have different ruby files under the same directory since that's how `rails` expects modules to be laid out. 

i.e.

folder `email_manager` -> module EmailModule

Really upto you but was a suggestion since it might make the code files not as long and have the unit tests short too

WDYT?

juliexxia: Ah yeah I was wondering about how I was going to mock that POST. Do you have a some time to hop on a call tomorrow? I'm not quite following what you're saying but I feel like you have the right instincts. I'm free whenever especially tomorrow morning!

spec/services/email_manager/donation_receipt_sender_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  let(:payment_intent) { create :payment_intent }
+  let(:receipt_params) do {
+      amount: 5000,
+      merchant: "Leaky Cauldron",
+      payment_intent: payment_intent,
+  }
+  end
+
+  it 'should call donation email sender methods' do
+    sender = class_double("EmailManager::Sender")
```
jtmckibb, line 13: this has some weird line breaks — mind running rubocop -a on your changes? non-blocking

PR #138, opened 2020-05-23
Add eager_load and cache_classes to staging.rb

Yong-L:
This is to ensure that environments are loaded up similarly to how it's
loaded up in production. Eager loading breaks if the modules are not
named properly and rails throws an exception

### DIFF
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -4,13 +4,14 @@ Rails.application.configure do
   # Settings specified here will take precedence over those in
   # config/application.rb.
 
-  # In the development environment your application's code is reloaded on
-  # every request. This slows down response time but is perfect for development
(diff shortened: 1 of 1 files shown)


config/environments/staging.rb
```diff
@@ -4,13 +4,14 @@
...
-  # Do not eager load code on boot.
-  config.eager_load = false
+  # Code is not reloaded between requests.
+  config.cache_classes = true
+
+  # Eager load code on boot. This eager loads most of Rails and
+  # your application in memory, allowing both threaded web servers
+  # and those relying on copy on write to perform better.
+  # Rake tasks automatically ignore this option for performance.
+  config.eager_load = true
```
jtmckibb, line 14: nice good find

PR #142, opened 2020-05-26
Order by amount_raised for pooling

jtmckibb:
When deciding how to split the extra cent, we should give it to the merchants with the least donations

### DIFF
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -125,7 +125,17 @@ class WebhooksController < ApplicationController
       if seller_id.eql?(Seller::POOL_DONATION_SELLER_ID)
         PoolDonationValidator.call({ type: item['item_type'] })
 
-        @donation_sellers = Seller.filter_by_accepts_donations
+        # TODO(jtmckibb): This is a very inefficient sort, since each time we
+        #                 call amount_raised, it has to fetch all of the
+        #                 associated Items, DonationDetails, and GiftCardDetails
+        #                 Then for each GiftCardDetail, it has to fetch every
+        #                 amount in an N+1 query, then sum everything.
+        #                 This all in an N log N sort is just a bad. Ideally,
+        #                 we would memoize amount_raised, and fix the N+1 query
+        #                 in GiftCardDetail that calculates amount.
+        @donation_sellers = Seller.filter_by_accepts_donations.sort_by do |s|
+          s.amount_raised
+        end
 
         # calculate amount per merchant
         # This will break if we ever have zero merchants but are still
--- a/spec/factories/contact.rb
+++ b/spec/factories/contact.rb
@@ -4,7 +4,5 @@ FactoryBot.define do
   factory :contact do
     email { Faker::Internet.email }
     is_subscribed { Faker::Boolean }
-
-    association :seller
   end
 end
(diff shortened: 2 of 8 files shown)


spec/factories/contact.rb
```diff
@@ -4,7 +4,5 @@
   factory :contact do
     email { Faker::Internet.email }
     is_subscribed { Faker::Boolean }
-
-    association :seller
```
jtmckibb, line 8: This association in the factory was automatically creating sellers when I created an Item via the factory. Since this factory association isn't really being used in any tests —  I'm going to remove it

app/controllers/webhooks_controller.rb
```diff
@@ -125,7 +125,17 @@ def handle_payment_intent_succeeded(
...
 
-        @donation_sellers = Seller.filter_by_accepts_donations
+        # TODO(jtmckibb): This is a very inefficient sort, since each time we
+        #                 call amount_raised, it has to fetch all of the
+        #                 associated Items, DonationDetails, and GiftCardDetails
+        #                 Then for each GiftCardDetail, it has to fetch every
+        #                 amount in an N+1 query, then sum everything.
+        #                 This all in an N log N sort is just a bad. Ideally,
+        #                 we would memoize amount_raised, and fix the N+1 query
+        #                 in GiftCardDetail that calculates amount.
```
jtmckibb, line 135: I want to fix this in a follow up since it doesn't *really* matter too much being that we only have 6 sellers. The real ideal would to be able to store this as a calculated column somehow and index on it

spec/requests/webhooks_spec.rb
```diff
@@ -158,21 +132,9 @@ def get_donation_detail(seller_id:)
...
-            expect(donation_to_seller1).not_to be_nil
-            expect(donation_to_seller2).not_to be_nil
-            expect(donation_to_seller2).not_to be_nil
-
-            expect(donation_to_seller1.amount).to eq(112)
-            expect(donation_to_seller2.amount).to eq(112)
-            expect(donation_to_seller3.amount).to eq(111)
+            expect(seller1.donation_amount).to eq(1_12)
+            expect(seller3.donation_amount).to eq(1_12)
+            expect(seller2.donation_amount).to eq(11_11)
```
jtmckibb, line 137: If we didn't have the sort on amount_raised, then it would have given the extra cent to sellers 1 and 2 because I think since the default sort is by id

spec/models/seller_spec.rb
```diff
@@ -139,6 +139,84 @@
...
+        item_donation1 = create(:item, seller: seller)
+        create(:donation_detail, item: item_donation1, amount: 200_00)
+
+        # Create a donation of $10
+        item_donation2 = create(:item, seller: seller)
+        create(:donation_detail, item: item_donation2, amount: 10_00)
+
+        # Create a donation of $50, refunded
+        item_donation3 = create(:item, seller: seller, refunded: true)
+        create(:donation_detail, item: item_donation3, amount: 50_00)
```
Yong-L, line 200: for the sake of keeping the unit test lean and mean, instantiation of all of these could probably move to a factory method

jtmckibb: Mind if I followed up? I copy-pasted these from the `sellers_helper_spec`

jtmckibb: but totes agree

PR #145, opened 2020-05-27
delete method for locations controller

ryannamgung:


app/controllers/gift_cards_controller.rb
```diff
@@ -19,6 +19,12 @@ def update
     json_response(item_gift_card_detail_json)
   end
 
+  # GET /gift_cards/show_all
```
rjvani, line 22: Let's remove this, and we can add an implementation after drafting the RFC.

jacoblb64:

PR #146, opened 2020-05-29
Email receipts for gift a meal

jtmckibb:
We emailing gifted mails to the disitributor, but the
purchaser never got a receipt. We plan to give the distributors
their gift cards manually from pulling from the db, so we
don't need to synchronously send them an email. We'll do
it manually. However, we do need to send an email to the
purchaser.


jtmckibb:
I also ran rubocop, so just review by commit

Issue #149, opened 2020-06-01
Duplicate email issue


solidiquis:
Note: Accidentally made this issue on the wrong repo earlier lol

The bug is that for “Gift A Meal” purchases, it breaks up the gift card purchase into a whole bunch of little gift cards attributed to a third party distributor and then sends a “thank you for your donation!” receipt to the purchaser but since the receipt sending is in the for loop that breaks up the gift cards, it ends up sending a whole bunch of tiny donation receipts to the purchaser.

So if you buy 5 meals for $25 (at $5 a meal), you’ll get 5 receipts for $5 each.

PR #150, opened 2020-06-01
Fix bug where DonationReceiptSender.call would fire for each line item

solidiquis:
Closes issue #149 

Issue: One email would send per line item for donation events.
Solution: Group line items by seller ID, sum the amounts of each group, the send one email per group.

I'm assuming that it's possible for an instance of PaymentIntent to have line items with multiple unique seller IDs.

Additionally, my hunch is that PoolDonationReceiptSender and GiftCardReceiptSender might have the same issue since they're fired within the for-loop; but I don't have as much context on those so I won't touch them.

Let me know if you would like me to refactor anything!

### DIFF
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -115,6 +115,8 @@ class WebhooksController < ApplicationController
 
     items = JSON.parse(payment_intent.line_items)
     recipient = payment_intent.recipient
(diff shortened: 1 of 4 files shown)


solidiquis:
Closing PR for now, going to wait for @Yong-L 's PR to get approved and merged first to avoid nasty code conflicts.

PR #154, opened 2020-06-11
Create gift cards task for seeding gift cards

Yong-L:
I made payment intent nullable to make it easier to seed gift cards


jtmckibb:
It might be worth adding tests for this in case the model changes

Yong-L:
> It might be worth adding tests for this in case the model changes

I added the test but I wasn't too sure about how to capture the output of the script... Also I have an `exit` in the condition which also exits the unit test so I'm not sure how to test against that

jtmckibb:
No worries as long as we make sure the gift cards are created ^^

PR #155, opened 2020-06-12
add migration for new delivery columns

juliexxia:


db/migrate/20200612212747_add_delivery_columns_to_seller.rb
```diff
@@ -0,0 +1,10 @@
+class AddDeliveryColumnsToSeller < ActiveRecord::Migration[6.0]
+  def change
+    enable_extension "hstore"
+
+    add_column :sellers, :delivery, :boolean
+    add_column :sellers, :delivery_options, :hstore
```
jtmckibb, line 6: Would it be better to create a new table called DeliveryPartner or DeliveryType? It would have name, icon, etc. associated. Then we could create another table called DeliveryOption which relates Seller to a DeliveryType. If we use an hstore the data isn't structured, and migrations could also be a problem if we want to add maybe an API key or something for example

juliexxia: Do things like name and Icon need to be stored on the backend? I thought it was only menu url and boolean for whether or not they deliver. I'll ping Olivia and check. 

Just so I'm understanding, is this the model you're thinking of:

DeliveryType
- `name` e.g. "grubhub"
- `icon` e.g. "some.url.to.image?"

DeliveryOption
- belongs_to Seller
- has_one DeliveryType
- url for web menu

Seller
- has_many DeliveryOptions 
- `delivers` (boolean)

If we didn't need to store stuff like name/icon on the backend, I guess it still makes sense to have a DeliveryOption table with a belongs_to/has_many relationship with Sellers. And deliverytype could be an enum.

juliexxia: What kind of problems do you envision with migrations for the hstore? And where in the process do you imagine we'd add an API key?

I guess the appeal of the hstore to me was exactly that data wasn't structured so that not all sellers had to know about it if a single seller got a new delivery type. But if delivery type is its own separate object or an enum it'll mostly function that way anyway and we'll have safety around not putting typo'd keys into an hstore

oliviasztanga: Ah! I see what he's saying and agree that may be a better approach. In that case, it would also make sense to store the icons: https://drive.google.com/drive/u/3/folders/1pCViVM9qddsBf1x5fXFUYEXLSsDfyyqY

Now I'm just unsure of the boolean to mark whether they have delivery/takeout available. The point of that boolean would be to determine whether or not to render the "Order Now" drop down. I could also just set up that logic to render based on whether there are any "DeliveryOption" entries, according to the approach Justin laid out.

The only issue is if the merchant is offering delivery/takeout by phone but not any other third party option. I think we need to store by phone as another "DeliveryType" and then "DeliveryOption," but then we don't need to store a URL -- I just need to pull the phone number we are already storing.

What do you think?

juliexxia: Cool let me think of an expanded structure and respond with it on this thread!

juliexxia: DeliveryType
- `name` e.g. "grubhub"
- `icon` e.g. "some.url.to.image?" `optional?`
*one instance is {name = "phone", icon = ?/null?}*

DeliveryOption
- `belongs_to` Seller
- `has_one` DeliveryType
- `url` e.g. "www.url.for/web/menu" `optional`
- `phone_number` e.g. "917-555-5555" 'optional'
*must have either url or phone number*
*if DeliveryType is phone, must have phone number*

^^ we can set up more tables to more cleanly enforce this second constraint, but I think that might be more complicated than it's worth. I'm not totally sure how model inheritance works in rails or if that even exists lol, but you could see something like a parent DeliveryOption class that is extended by ThirdPartyDeliveryOption and PhoneDeliveryOption which have a url and a phone number respectively. 

Seller
- `has_many` DeliveryOptions `optional`

I think if it's straightforward for you (Olivia (: ) to set up logic to query whether a seller's DeliveryOptions reference is empty or not, then we should drop the `has_delivery` boolean. If not, I think we should have it for ease of access and we don't lose that much by way of duplication.

jtmckibb: You can augment the Seller to make it return an array of DeliveryOptions in the SellersHelper! Similar to locations

jtmckibb: But to answer your questions about problems I saw:

- What if one day we want to support ordering from our website. I could imagine we'd need to store a "merchant token" on this object
- What if we one day want to store things like colors associated to determine something like hover state with each DeliveryType eg) GrubHub would be red/white, and Caviar would be orange/white, etc. A map would only allow us to store a single value per key.

Just in general I think it's better to have a model that is extensible in the case we need to add more than just a URL to a DeliveryOption.

oliviasztanga: Yeah, I agree with dropping the `has_delivery` boolean.

Speaking of locations, how would we handle this if a seller has multiple locations and multiple phone numbers?

jtmckibb:
We haven’t supported multiple locations so far. Let’s just assume they have
one. I’d even venture to say to update the API to only return the first
location as a follow up.

db/migrate/20200612212747_add_delivery_columns_to_seller.rb
juliexxia: I believe if we implement as suggest, If there are multiple locations a seller should be able to have multiple DeliveryOptions that are of the phone DeliveryType

juliexxia: Also just seeing Justin's comment, that works too

justintmckibben: I'd use menu_url, and icon_url or logo_url. Also as for phone number, would this be different from the phone number on the location? But yes! The model looks good to me :)

juliexxia:
Closing in favor of #160

PR #158, opened 2020-06-14
Open Hours Backend

Jeremy-Xue:
Relevant files are open_hour.rb and open_hour_controller.rb

Other files were changes from running rubocop -a. Had some issues rebasing to clean up the commit history but can take another try at it if needed

Heavily based this off of the implementation for menu_items since it seems to be very similar in functionality. 

Open hours are associated with each seller, and contain fields for day of the week (1-7) and open and close times (time datatype).

Does not account for holidays (can add text on frontend).

### DIFF
--- a/app/controllers/open_hour_controller.rb
+++ b/app/controllers/open_hour_controller.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+class OpenHourController < ApplicationController
+  before_action :set_seller
+  before_action :set_seller_open_hour, only: %i[update destroy]
+
+  def index
+    json_response(@seller.open_hour)
+    end
+
+  def create
+    json_response(@seller.open_hour.create!(create_open_hour_params), :created)
+  end
+
+  def update
+    @open_hour.update(update_open_hour_params)
+    @open_hour.save
+    json_response(@open_hour)
+  end
+
+  def destroy
+    @open_hour.destroy
+
+    head :no_content
+  end
+
+  private
+
+  def create_open_hour_params
+    params.require(:seller_id)
+    update_params
+  end
+
+  def update_open_hour_params
+    params.require(:seller_id)
+    params.require(:id)
+    update_params
+  end
+
+  def update_params
+    params.permit(
+      :day,
+      :open,
+      :close
+    )
+  end
+
+  def set_seller
+    @seller = Seller.find_by!(seller_id: params[:seller_id])
+  end
+
+  def set_seller_menu_item
+    @open_hour = @seller.open_hour.find_by!(id: params[:id]) if @seller
+  end
+end
--- a/app/models/open_hour.rb
+++ b/app/models/open_hour.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+# == Schema Information
+#
+# Table name: open_hours
+#
+#  id         :bigint           not null, primary key
+#  close      :time
+#  day        :integer
+#  open       :time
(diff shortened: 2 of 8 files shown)


app/controllers/open_hour_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+    json_response(@seller.open_hour)
+    end
+
+  def create
+    json_response(@seller.open_hour.create!(create_open_hour_params), :created)
+  end
+
+  def update
+    @open_hour.update(update_open_hour_params)
+    @open_hour.save
```
jtmckibb, line 17: I don't think you need this save. Do the tests still pass without it?

app/models/open_hour.rb
```diff
@@ -0,0 +1,37 @@
...
+#
+# Foreign Keys
+#
+#  fk_rails_...  (seller_id => sellers.id)
+#
+class OpenHour < ApplicationRecord
+  belongs_to :seller
+  validates_presence_of :day, :close, :open
+  # 1:Monday - 7:Sunday
+  validates_inclusion_of :day, in: 1..7
```
jtmckibb, line 27: Would it be possible to make this 0..6? This way we can X mod 7 vs (X mod 7) + 1. In general starting a zero is just going to be more natural for programmers

app/models/open_hour.rb
```diff
@@ -0,0 +1,37 @@
...
+#
+# Indexes
+#
+#  index_open_hours_on_seller_id  (seller_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (seller_id => sellers.id)
+#
+class OpenHour < ApplicationRecord
```
jtmckibb, line 23: Can you add an enum to map MON to 0, TUE to 1, etc. to match the Square API: https://developer.squareup.com/reference/square/objects/BusinessHoursPeriod#definition__property-day_of_week

enums: https://naturaily.com/blog/ruby-on-rails-enum
https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html

app/models/open_hour.rb
```diff
@@ -0,0 +1,37 @@
...
+  belongs_to :seller
+  validates_presence_of :day, :close, :open
+  # 1:Monday - 7:Sunday
+  validates_inclusion_of :day, in: 1..7
+  validate :opens_before_closes
+
+  protected
+
+  def opens_before_closes
```
jtmckibb, line 32: This doesn't take into account if you open on a Monday, and close at 2am on a Tuesday

spec/requests/open_hour_request_spec.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'OpenHours', type: :request do
```
jtmckibb, line 5: Mind adding GET, POST, PUT and DELETE tests here?

spec/models/open_hour_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+#
+#  index_open_hours_on_seller_id  (seller_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (seller_id => sellers.id)
+#
+require 'rails_helper'
+
+RSpec.describe OpenHour, type: :model do
```
jtmckibb, line 25: Could you add tests for:

- validates_inclusion_of
- enum
- opens_before_closes
- validates_presence_of

PR #160, opened 2020-06-17
Juliexxia/delivery

juliexxia:
API CHANGES:

DeliveryType
- `name` e.g. "grubhub"
- `icon` e.g. "some.url.to.image?" `optional?`
*one instance is {name = "phone", icon = ?/null?}*

DeliveryOption
- `belongs_to` Seller
- `has_one` DeliveryType
- `url` e.g. "www.url.for/web/menu" `optional`
- `phone_number` e.g. "917-555-5555" 'optional'
*must have either url or phone number*
*if DeliveryType is phone, must have phone number*

^^ we can set up more tables to more cleanly enforce this second constraint, but I think that might be more complicated than it's worth. I'm not totally sure how model inheritance works in rails or if that even exists lol, but you could see something like a parent DeliveryOption class that is extended by ThirdPartyDeliveryOption and PhoneDeliveryOption which have a url and a phone number respectively. 

Seller
- `has_many` DeliveryOptions `optional`


app/models/delivery_option.rb
```diff
@@ -0,0 +1,15 @@
+# == Schema Information
+#
+# Table name: delivery_options
+#
+#  id           :bigint           not null, primary key
+#  url          :string
```
jacoblb64, line 6: I'm trying to think of any other kinds of data we might need to store here. Possibly for mobile, would we want to have a deep link to directly open in an app?

db/migrate/20200617023402_add_delivery_type_key_to_delivery_option.rb
```diff
@@ -0,0 +1,6 @@
+class AddDeliveryTypeKeyToDeliveryOption < ActiveRecord::Migration[6.0]
+  def change
+  	# This should have been :delivery_option
+  	add_reference :delivery_types, :delivery_options, index:true
```
Yong-L, line 4: Just gotta add `foreign_key: true` here

PR #162, opened 2020-06-17
Multiple Emails Bug

### LINKED ISSUE #149: Duplicate email issue

solidiquis:
Closes #149

### DIFF
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -115,6 +115,8 @@ class WebhooksController < ApplicationController
 
     items = JSON.parse(payment_intent.line_items)
     recipient = payment_intent.recipient
+    is_donation = nil
+
     items.each do |item_json|
       # TODO(jtmckibb): Add some tracking that tracks if it breaks somewhere
       # here
@@ -139,19 +141,14 @@ class WebhooksController < ApplicationController
         merchant_name = Seller.find_by(seller_id: seller_id).name
         case item_json['item_type']
         when 'donation'
+          is_donation ||= true
+
           WebhookManager::DonationCreator.call({
                                                  seller_id: seller_id,
                                                  payment_intent: payment_intent,
                                                  amount: amount
                                                })
-          EmailManager::DonationReceiptSender.call({
-                                                     payment_intent: payment_intent,
-                                                     amount: amount,
-                                                     merchant: merchant_name,
-                                                     email: payment_intent.purchaser.email
-                                                   })
         when 'gift_card'
-
           is_distribution = item_json['is_distribution']
 
           gift_card_detail = WebhookManager::GiftCardCreator.call({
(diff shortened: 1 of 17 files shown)


app/controllers/webhooks_controller.rb
```diff
@@ -185,5 +182,17 @@ def handle_payment_intent_succeeded(
         end
       end
     end
+
+    if is_donation
+      # Send separate email for each seller.
+      grouped_items = items.group_by { |li| li["seller_id"] }
```
justintmckibben, line 188: nit, mind running rubocop -a

app/controllers/webhooks_controller.rb
```diff
@@ -139,19 +141,14 @@ def handle_payment_intent_succeeded(
         merchant_name = Seller.find_by(seller_id: seller_id).name
         case item_json['item_type']
         when 'donation'
+          is_donation ||= true
```
justintmckibben, line 144: wow what is this ruby madness lol

spec/controllers/webhooks_controller_spec.rb
```diff
@@ -3,4 +3,48 @@
 require 'rails_helper'
 
 RSpec.describe WebhooksController, type: :controller do
+  describe "POST create" do
+    let!(:payment_intent) do
+      create :payment_intent, :with_line_items, square_payment_id: 'square-id', square_location_id: "OIJWEOIFJEOIFJ"
```
justintmckibben, line 8: could you use ENV['SQUARE_LOCATION_ID']?

app/controllers/webhooks_controller.rb
jacoblb64, line 144: Maybe this is a ruby quirk i’m not familiar with, but why are we setting this to nil initially instead of false, then `||=`’ing it to true instead of just setting it?

solidiquis: Because if I initially set it to false, then set it to true in the loop without using the conditional assignment operator, is_donation will keep getting re-assigned to true after every loop, which is like negligibly more time expensive, but still haha. I COULD use an if-statement but would look kind of ugly haha. Open to discussion though!

Screen Shot 2020-06-17 at 3 58 15 PM

jacoblb64: Interesting, makes sense that `||=` is faster than a raw assignment. Where’s the advantage come from the initial `nil` vs `false` though?

solidiquis: Ahhh I had assumed that if a = false then a ||= true would still give a = false but turns out that isn't the case haha. I'll change that. Thanks man!

PR #163, opened 2020-06-20
Add action controller caching to staging

Yong-L:
Temporarily checking if caching makes the requests faster in staging environment before pushing to prod for testing.


Yong-L:
Followed https://devcenter.heroku.com/articles/memcachier#rails

PR #166, opened 2020-06-27
Add tests and fixes to original hours backend PR

Jeremy-Xue:


db/schema.rb
```diff
@@ -115,6 +115,17 @@
     t.index ["seller_id"], name: "index_menu_items_on_seller_id"
   end
 
+  create_table "open_hours", force: :cascade do |t|
+    t.bigint "seller_id", null: false
+    t.time "open"
+    t.time "close"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+    t.integer "openday"
```
justintmckibben, line 124: nit and you don't have to change it if you don't want to / can do it as a follow up; but could you make this open_day?

db/schema.rb
```diff
@@ -115,6 +115,17 @@
     t.index ["seller_id"], name: "index_menu_items_on_seller_id"
   end
 
+  create_table "open_hours", force: :cascade do |t|
```
justintmckibben, line 118: Curious, how would you model a store that's open 24-hours?

app/models/open_hour.rb
```diff
@@ -0,0 +1,41 @@
...
+  enum days: { MON: 0, TUE: 1, WED: 2, THU: 3, FRI: 4, SAT: 5, SUN: 6 }
+  enum openday: { MON: 0, TUE: 1, WED: 2, THU: 3, FRI: 4, SAT: 5, SUN: 6 }, _suffix: true
+  enum closeday: { MON: 0, TUE: 1, WED: 2, THU: 3, FRI: 4, SAT: 5, SUN: 6 }, _suffix: true
+  validates_inclusion_of :openday, in: days.keys
+  validates_inclusion_of :closeday, in: days.keys
+  validate :opens_before_closes
+
+  protected
+
+  def opens_before_closes
```
justintmckibben, line 36: Do you have a test for this?

Jeremy-Xue: Added!

db/schema.rb
Jeremy-Xue: changed!

db/schema.rb
Jeremy-Xue: I think we could store it as 12am (day n) to 12am (day n+1) and have a check on the frontend to show "Open 24 Hours"

PR #167, opened 2020-06-27
Delegating GAM Payments to Non-Profit Square accounts (Think!Chinatown)

jacoblb64:
For Think!Chinatown (and hopefully future partners), we have to delegate payments to their square accounts instead of ours for non-profit type transactions. Right now this is just GAM, but we want to make this generic. The current model as of now is to use one "Send Chinatown Love" Location ID within Think!Chinatown's square account, but we could support other scenarios too.

Overview:
- Adding a `non_profit_location_id` to the sellers model, which represents an alternative square location id to use for non-profit type transactions.
- If performing a GAM transaction (`is_distribution` is `true`), and the seller has a valid `non_profit_location_id` set, then use that location id instead of the seller's usual one
- Based on that we have to use the right Square credentials for performing the transaction

Supporting details:
- Adding more env variables with all the new Think!Chinatown credentials
- Making sure we support webhooks from their square account
- Appropriate schema and migration for the new field
- An new seller to be seeded for testing that supports these scenarios
- A more than healthy dose of testing!

Confirmed working payment in Think!Chinatown's sandbox app!

Square credit card nonces have to be encoded from the same account, so we have corresponding front-end changes [in this PR (!207)](https://github.com/sendchinatownlove/sendchinatownlove.github.io/pull/207)


app/services/square_manager/payment_creator.rb
```diff
@@ -32,5 +32,14 @@ def create_payment_body
         location_id: location_id
       }
     end
+
+    def which_access_token_from_location_id(location_id:)
```
justintmckibben, line 36: nit `access_token_from_location_id`

app/controllers/charges_controller.rb
```diff
@@ -113,7 +113,11 @@ def validate(seller_id:, line_items:)
   def create_square_payment_request(
     nonce:, amount:, email:, name:, seller:, line_items:, is_distribution:
   )
-    square_location_id = seller.square_location_id
+    square_location_id = if is_distribution && !seller.non_profit_location_id.blank?
```
justintmckibben, line 116: is !.blank? the same as .present?

db/seeds.rb
```diff
@@ -27,6 +27,28 @@
     # https://docs.google.com/document/d/1UPNCwjWS_T7XT5AXsewphu6NvNdV7TQLSJub-RBRAG0/edit
     square_location_id: ENV['SQUARE_LOCATION_ID']
   },
+  {
+    seller_id: '46-mott',
+    cuisine_name: 'Bakery',
+    name: '46 Mott',
+    story: 'Patrick is awesome, tbh',
```
justintmckibben, line 34: true story

spec/controllers/charges_controller_spec.rb
```diff
@@ -3,4 +3,78 @@
 require 'rails_helper'
 
 RSpec.describe ChargesController, type: :controller do
+  include MockApiResponseHelper
+
+  describe "POST create" do
```
justintmckibben, line 8: nit single quotes instead of double. Can you run `rubocop -a`

app/services/square_manager/payment_creator.rb
justintmckibben: is this codepath tested any way?

justintmckibben:
Looks great! Do you mind squashing the commits before merging?

app/services/square_manager/payment_creator.rb
jacoblb64: sure changed the name, and I don't think it's actually being hit in unit tests, let me see if I can whip one up

app/controllers/charges_controller.rb
jacoblb64: ahah realized that it is! switched to .present? as I think the positive case makes more sense.

spec/controllers/charges_controller_spec.rb
jacoblb64: 🏆

jacoblb64:
> Looks great! Do you mind squashing the commits before merging?

I usually choose the "squash and merge" option when merging a PR. Is that sufficient, or do we usually rebase/squash on the PR itself?

app/services/square_manager/payment_creator.rb
jacoblb64: Just added a unit test for this which I'm happy about 👌

justintmckibben:
Squash and merge is fine!

PR #168, opened 2020-07-03
Task to send emails of recent gift cards to merchants

jacoblb64:
Working and ready for review!

This is a rake task to email a given seller (and email) vouchers for their businesses over a certain range of time. The default is the last year, which is pretty much "all time" for us. The time is an argument to the task if we want it.

I think this will be short-lived as @justintmckibben is taking this query and building out an awesome endpoint to build a persistent, dedicated front-end site to host this info for merchants to reference. (https://github.com/sendchinatownlove/ruby/pull/175)

As this is probably gonna mostly be a one-off, I've kept it light on the formatting / styling with just a plain email.

Screen Shot 2020-07-11 at 9 33 17 PM

### DIFF
--- a/lib/tasks/emailer.rake
+++ b/lib/tasks/emailer.rake
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+namespace :emailer do
+  desc 'email task runner'
+  task vouchers_to_merchants: :environment do
+    desc 'sends all voucher codes to all of their merchants'
+    # c = Contact.find_by(seller_id: 3)
+    # puts(c.email)
+
+    sql = 'select gcd.gift_card_id,
+        gca.value,
+        c_s.name as seller_name,
+        c_s.email as seller_email,
+        c_r.email as recipient_email,
+        c_r.name as recipient_name,
+        gcd.created_at,
+        gcd.expiration
+        from gift_card_details gcd
+        join (
+            select distinct on (gift_card_detail_id) *
+            from gift_card_amounts
+            ) as gca on gca.gift_card_detail_id = gcd.id
+        join items i on gcd.item_id = i.id
+        join contacts c_s on i.seller_id = c_s.seller_id
+        join contacts c_r on gcd.recipient_id = c_r.id
+        where gcd.created_at > current_date - 30'
+
+    # records_array = ActiveRecord::Base.connection.execute(sql)
+    # puts(records_array.inspect)
+
+    r = GiftCardDetail.connection.select_all(sql)
+
+    html = ''
+    html += '<style> table, tr, td {border: 1px solid black}</style>'
+    html += <table><tr>'
+
+    r.columns.each do |col|
+      html += '<th>' + col + '</th>'
+    end
+
+    html += '</tr>'
+
+    r.each do |a|
+      html += '<tr>'
+      a.each do |_key, val|
+        html += '<td>' + val.to_s + '</td>'
+      end
+      html += '</tr>'
+    end
+
+    html += '</table>'
+
+    puts(html)
+
+    puts('sending...')
+    EmailManager::Sender.send_receipt(to: ENV['EMAIL_ADDRESS'], html: html)
+    puts('sent?')
+  end
+end
--- a/lib/tasks/emailer.rake
+++ b/lib/tasks/emailer.rake
@@ -76,7 +76,7 @@ namespace :emailer do
     end
 
     html += '</table>'\
-            '<p>For questions about how Send Chinatown Love Gift Cards work, '\
+            '<p>For questions about how Send Chinatown Love vouchers work, '\
             'please feel free to reply to this email!</p>'\
             '<p>All the best from the team at Send Chinatown Love</p>'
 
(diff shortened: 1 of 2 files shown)


lib/tasks/emailer.rake
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+namespace :emailer do
+  desc 'email task runner'
+  task vouchers_to_merchants: :environment do
+    desc 'sends all voucher codes to all of their merchants'
+
+    s_sql = 'select c.email as seller_contact_email,
```
justintmckibben, line 8: oh wow any reason why you used sql instead of ActiveRecord?

jacoblb64: This one could most likely be converted to ActiveRecord--I started with the other query first and it seemed more complex than I could get through ActiveRecord, but maybe I'll give it a shot. I'm not the hugest fan of all this raw sql here, but it was a good place for me to start as I figure I probably know that better than the ActiveRecord API 😬

justintmckibben: ah makes sense. active record is nice because it has checks and stuff built into the models. happy to pair with you on this!

lib/tasks/emailer.rake
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+namespace :emailer do
+  desc 'email task runner'
+  task vouchers_to_merchants: :environment do
+    desc 'sends all voucher codes to all of their merchants'
```
justintmckibben, line 6: currently it sends every gift card they have — i wonder if that's what they want, or do they actually just want the ones within the last week

we should ask chloe and marcia what they think the merchants would prefer. it's worth mentioning that sending them everything would include updated values

lib/tasks/emailer.rake
jacoblb64: Sounds good to me! Let me take another pass at it, then would be great to pair sometime this weekend!

lib/tasks/emailer.rake
jacoblb64: Gotchya, I was imagining we would send them any gift card updated within the last X number of days/weeks, but I suppose maybe it's easier to send them a list of all gift cards every time and they can just toss the last list?

Would be good to hear what they might prefer

justintmckibben: yeah it's important to note that the values on our dbs are the customer's values, while the values on their paper is the true value — so they'd have to "merge" the lists every time if they did that. we def need seller input here

lib/tasks/emailer.rake
```diff
@@ -0,0 +1,75 @@
...
+
+      Rails.logger.info('Getting gift cards for ' + seller.seller_id)
+      html = '' # start building html with an empty string
+
+      html += "<h3> Hello #{seller.distributor.name} at #{seller.seller_id} </h3>"
+      html += "<p>Here's last #{_args.time_range}'s vouchers:</p>"
+
+      # query for all gift cards info for that seller
+      query = GiftCardDetail
+              .select(:gift_card_id, :value, :name, :email, :created_at, :expiration)
```
justintmckibben, line 33: I think you want seller_gift_card_id instead of gift_card_id

jacoblb64: Thanks for the catch!

lib/tasks/emailer.rake
```diff
@@ -0,0 +1,75 @@
...
+          html += '<td>' + val.to_s + '</td>'
+        end
+        html += '</tr>'
+      end
+
+      html += '</table>'
+      html += '<p>All the best from the team at Send Chinatown Love</p>'
+
+      Rails.logger.info('Sending vouchers to ' + seller.distributor.email)
+      EmailManager::Sender.send_receipt(to: ENV['EMAIL_ADDRESS'], html: html)
```
justintmckibben, line 72: We should add a column to the Seller for email

Actually I think we should do a relationship to Contact called "owner" and add in Seller names

jacoblb64: woops this should be `seller.distributor.email` thanks for the catch! This was for testing 😬

jacoblb64: After conversing offline, we determined that it's best for now to take this as an argument

lib/tasks/emailer.rake
```diff
@@ -0,0 +1,86 @@
...
+        val = '$' + (val / 100).to_s if key == 'value'
+        val = val ? 'Y' : 'N' if key == 'single_use'
+        val = 'N/A' if val.blank?
+        html += '<td>' + val.to_s + '</td>'
+      end
+      html += '</tr>'
+    end
+
+    html += '</table>'\
+            '<p>For questions about how Send Chinatown Love Gift Cards work, '\
```
justintmckibben, line 79: vouchers*

lib/tasks/emailer.rake
```diff
@@ -0,0 +1,86 @@
...
+            'please feel free to reply to this email!</p>'\
```
justintmckibben, line 80: I need to add receipts@sendchinatownlove.com to our domain

lib/tasks/emailer.rake
jacoblb64: thanks for the catch!

lib/tasks/emailer.rake
jacoblb64: Makes sense, let's be sure to do that before we actually send this out 👍

PR #170, opened 2020-07-05
Add fees to gift a meal for Think!Chinatown sponsorship

justintmckibben:
We created a new model called "Fee" which allows for
us to apply a fee to the "total amount" it costs per
meal. eg) a $10 meal with a 10% fee would cost $11,
but would generate a $10 gift card. The reasoning for
something like this is to set aside that extra 10% for
things like processing fees or fiscal sponsorships.

This is important for our Think!Chinatown sponsorship
since donations that go through them are tax deductable.


app/controllers/fees_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  # GET /fees
+  def index
+    json_response(Fee.all)
+  end
+
+  # POST /fees
+  def create
+    puts create_params
+    fee = Fee.create!(create_params)
```
danthaman44, line 14: should this be @fee? Im not an expert on ruby syntax so plz disregard if not the case

app/models/fee.rb
```diff
@@ -0,0 +1,18 @@
...
+#  active     :boolean          default(TRUE)
+#  multiplier :decimal(, )      default(1.0)
+#  seller_id  :bigint
+#
+# Indexes
+#
+#  index_fees_on_seller_id  (seller_id)
+#
+class Fee < ApplicationRecord
+  belongs_to :seller
```
danthaman44, line 17: should also update Seller to `has_one` Fee?

app/controllers/fees_controller.rb
justintmckibben: @fee is the same thing as this.fee in Java, while fee without the @ is creating a local variable. In this case, it doesn't matter since nobody accesses this.fee after the call of create — so either would work

app/models/fee.rb
justintmckibben: Good catch, will add!

app/controllers/fees_controller.rb
jacoblb64: guys stop tagging this guy Joe 😂

app/helpers/sellers_helper.rb
```diff
@@ -4,9 +4,20 @@ module SellersHelper
...
+    fees = seller.fees
+
     json = seller.as_json
     json['distributor'] = distributor.as_json unless distributor.nil?
     json['locations'] = locations.as_json
+    json['fees'] = fees.as_json
+
+    # Take into account the associated fees when calculating the cost per meal
+    # for the customer. eg) We might generate a $10 gift card for GAM, but if
+    # there is a 10% fee associated with creating that gift card, then we should
```
jacoblb64, line 16: Question about flow: So I'm guessing that we'll be displaying the fee amount on the front-end to the user--would the charged amount be the with-fee or without-fee amount?

If it's the with-fee amount, then do we need to do this math here, or should we be doing it on the front-end?

db/schema.rb
```diff
@@ -56,6 +56,13 @@
     t.index ["idempotency_key", "event_type"], name: "index_existing_events_on_idempotency_key_and_event_type", unique: true
   end
 
+  create_table "fees", force: :cascade do |t|
```
jacoblb64, line 59: Would you mind also adding some fees to the seed file? Will also be helpful for testing.

app/controllers/fees_controller.rb
justintmckibben: LOL

app/controllers/fees_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+  def create_params
+    params.require(:seller_id)
+    update_params
+  end
+
+  def update_params
+    # The multiplier or other attributes such as that of a fee should be
+    # immutable. If you need to modify a fee, set it to inactive and create
+    # a new one. Modifying a fee's multiplier with associated PaymentIntents
+    # could have the side-effect of "re-writing history," changing the fees
```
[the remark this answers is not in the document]
justintmckibben: I thought about that. I can serialize the fees, etc. as a JSON and store it on the PaymentIntent when it's created. It is definitely worth pointing out that there is a difference from a "HistoricalFee" and a "CurrentFee". With this model, I am conflating the two. 

The advantage of having it as a separate object as opposed to a serialized object is that we can find all PaymentIntents that were charged with a particular version of a fee. If we serialized it and stored it on the table, it would have to do an N search to find all PaymentIntents that used a particular version of a fee, since there would be no index on that column.

app/helpers/sellers_helper.rb
justintmckibben: Good question. I prefer making the frontend dumb, and doing the math on the backend to reduce potential errors. As for whether or not we're displaying the fee amount on the frontend, I'm not sure. I think that to the user, as long as they know that their $11 gets a meal to the end user, that's all that matters to them. The details of the fact that 2.75% goes to cc processing, and 7.25% goes to Think!Chinatown — might just confuse the end user. 

What do you think as a user? Also @nanxiy might have an opinion here too

db/schema.rb
justintmckibben: Great idea!

app/models/fee.rb
justintmckibben: I added this but by nature it only "has one" fee since the seller_id is on the fee's table. belongs_to is sufficient enough in this case. has_one is only useful if fee_id existed on the sellers table.

app/helpers/sellers_helper.rb
justintmckibben: Bumping, @nanxiy do you think we *need* to show the fees charged per meal? or is it cool to keep it simple and just say that $11 buys a meal! (even if that meal is a $10 meal)

app/controllers/fees_controller.rb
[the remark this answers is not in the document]
justintmckibben: I kept it as is, but open to discussing this approach!

app/controllers/fees_controller.rb
justintmckibben: Sorry Joe! Resolving this comment 🙃

justintmckibben:
I just thought about this — should build fees such that < 1 it's a discount, and > 1 it costs more? So if a multiplier is .1 on a $10 meal it'll be $1, and if it's 1.1 on a $10 meal it'll be $11?

Right now it's built such that it only can add cost to a meal and can't take away. 

What do y'all think @danthaman44 @jacoblb64

app/helpers/sellers_helper.rb
```diff
@@ -4,9 +4,22 @@ module SellersHelper
...
+    # charge the customer $11.
+    if seller.cost_per_meal.present?
+      json['cost_per_meal'] = seller.cost_per_meal + fees.inject(0) do |_total, fee|
+        (seller.cost_per_meal * fee.multiplier).ceil
+      end
+    end
```
justintmckibben, line 22: I just thought about this — should build fees such that < 1 it's a discount, and > 1 it costs more? So if a multiplier is .1 on a $10 meal it'll be $1, and if it's 1.1 on a $10 meal it'll be $11?

Right now it's built such that it only can add cost to a meal and can't take away.

What do y'all think @danthaman44 @jacoblb64

justintmckibben: Actually, I think I might keep it as is. If we want to have discounts, I think we can model them as a separate object called Discounts. It just makes it clearer that way

PR #172, opened 2020-07-09
update to only return first location

ath-chen:


app/helpers/sellers_helper.rb
```diff
@@ -2,7 +2,7 @@
 
 module SellersHelper
   def self.generate_seller_json(seller:)
-    locations = seller.locations
+    locations = seller.locations[0]
```
justintmckibben, line 5: nit: seller.locations.first

justintmckibben, line 5: also do you mind logging Rails.logger.info whenever seller.locations.size is > 1? It'll be useful information for us to see when a seller has more than one location so that we can go and delete the extra location

app/helpers/sellers_helper.rb
```diff
@@ -2,7 +2,7 @@
 
 module SellersHelper
```
justintmckibben, line 3: This module has tests that you're going to have to update — also I have a feeling that some other adjacent tests will fail 😅

app/helpers/sellers_helper.rb
danthaman44, line 5: lets rename this variable 'location'. Ideally we would change the name of the field in the JSON to 'location' as well, but that will require a simultaneous front-end change

justintmckibben: Dan brings up a good point. We should actually probably keep this field as-is, and add a new field called "location" that only returns the first location. Then we should migrate the frontend to use the new field, THEN delete the old "locations" field

app/helpers/sellers_helper.rb
ath-chen: i checked! didn't see anything and ctrl + f didn't see any location test errors besides the one in the helper

app/helpers/sellers_helper.rb
ath-chen: made this and all the other VVV updates

jacoblb64:
Closing this as I don't think it's relevant anymore -- our front-end handles and expects multiple locations at this point. @ath-chen feel free to re-open or discuss if you feel otherwise!

PR #173, opened 2020-07-10
Reducing SQL queries by using distinct on

jacoblb64:


app/models/seller.rb
```diff
@@ -75,26 +75,25 @@ def amount_raised
...
-                  .inject(0) do |sum, gift_card|
-      sum + gift_card.amount
-    end
+    GiftCardDetail
+      .joins(:item)
+      .where(items: {
+               seller_id: id,
+               refunded: false
+             })
+      .joins("join (#{GiftCardAmount.latest_amounts_sql}) as la on la.gift_card_detail_id = gift_card_details.id")
```
justintmckibben, line 84: I didn't know that you could do this

app/models/seller.rb
```diff
@@ -75,26 +75,25 @@ def amount_raised
...
-                  .inject(0) do |sum, donation|
-      sum + donation.amount
-    end
+    DonationDetail
+      .joins(:item)
+      .where(items: {
+               seller_id: id,
+               refunded: false
+             })
+      .sum(:amount)
```
justintmckibben, line 96: Does this sum in SQL as opposed to in rails?

justintmckibben:
Still looks good!

spec/models/seller_spec.rb
```diff
@@ -180,6 +180,18 @@
           gift_card_detail: gift_card_detail2,
           created_at: Time.current + 1.day
         )
+        # Extraneous gift card amounts that should be ignored since it only
+        # should use the most recent ammount (aka the one updated a day later)
```
stzhng, line 184: amount typo; rest looks great to me

justintmckibben: Will fix in a followup!

app/models/seller.rb
jacoblb64: Yup! Makes it into `SELECT SUM(amount)`--which I believe should be more efficient here as the query is now retreiving only one value, instead of the set which we then have to sum in code

app/models/seller.rb
jacoblb64: I just learned it too ahaha, seemed like the best way to do this!

PR #175, opened 2020-07-11
Add seller gift cards endpoint

justintmckibben:
Created a new endpoint that lists all of the gift cards for a Seller given a secret access token. We're going to use this endpoint to allow for Sellers to look up their gift cards live.

### DIFF
--- a/spec/factories/gift_card_detail.rb
+++ b/spec/factories/gift_card_detail.rb
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :gift_card_detail do
     gift_card_id { Faker::Alphanumeric.alphanumeric(number: 64) }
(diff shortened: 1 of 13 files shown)


spec/factories/gift_card_detail.rb
```diff
@@ -3,7 +3,7 @@
 FactoryBot.define do
   factory :gift_card_detail do
     gift_card_id { Faker::Alphanumeric.alphanumeric(number: 64) }
-    seller_gift_card_id { Faker::Alphanumeric.alphanumeric(number: 64) }
+    seller_gift_card_id { '#' + Faker::Alphanumeric.alphanumeric(number: 5) }
```
jacoblb64, line 6: should we update the seeds to match?

jacoblb64:
@justintmckibben can we merge this?

PR #177, opened 2020-07-17
Add divisibility validation for charged amounts and seller cost per meals

rjvani:
For the Gift A Meal flow, check that the line item `amount`s are divisible by the seller's `cost_per_meal`.


app/controllers/charges_controller.rb
```diff
@@ -77,8 +77,9 @@ def charge_params
     )
   end
 
-  def validate(seller_id:, line_items:)
-    unless Seller.find_by(seller_id: seller_id).present?
+  def validate(seller_id:, line_items:, is_distribution:)
+    seller = Seller.find_by(seller_id: seller_id)
+    unless seller.present?
```
justintmckibben, line 82: I'm surprised this didn't already exist

PR #178, opened 2020-07-19
Add controller for delivery_options

juliexxia:
Heavily modeled off of https://github.com/sendchinatownlove/ruby/pull/27/files

Question - do I need to add any routes or tests for the fact that each DeliveryOption has a DeliveryType?

### DIFF
--- a/app/controllers/delivery_options_controller.rb
+++ b/app/controllers/delivery_options_controller.rb
@@ -11,7 +11,7 @@ class DeliveryOptionsController < ApplicationController
       json
     end
 
-    json_response(delivery_options)
+    json_response(delivery_options.sort_by{|de| de['delivery_type']['name'].to_i})
   end
 
   # POST /sellers/:seller_id/delivery_options
--- a/app/controllers/delivery_options_controller.rb
+++ b/app/controllers/delivery_options_controller.rb
@@ -4,7 +4,14 @@ class DeliveryOptionsController < ApplicationController
 
   # GET /sellers/:seller_id/delivery_options
   def index
-    json_response(@seller.delivery_options)
+    delivery_options = @seller.delivery_options.map do |delivery_option|
+      delivery_type = delivery_option.delivery_type
+      json = delivery_option.as_json
+      json['delivery_type'] = delivery_type.as_json
+      json
+    end
+
+    json_response(delivery_options)
   end
 
   # POST /sellers/:seller_id/delivery_options
--- a/app/controllers/delivery_options_controller.rb
+++ b/app/controllers/delivery_options_controller.rb
@@ -7,7 +7,7 @@ class DeliveryOptionsController < ApplicationController
     json_response(@seller.delivery_options)
   end
 
-  # POST /sellers/:seller_id/menu_items
+  # POST /sellers/:seller_id/delivery_options
   def create
     json_response(@seller.delivery_options.create!(create_delivery_option_params), :created)
   end
--- a/app/controllers/delivery_options_controller.rb
+++ b/app/controllers/delivery_options_controller.rb
@@ -11,7 +11,9 @@ class DeliveryOptionsController < ApplicationController
       json
     end
 
-    json_response(delivery_options.sort_by{|de| de['delivery_type']['name'].to_i})
+    Rails.logger.info delivery_options
+    json_response(delivery_options)
+#     json_response(delivery_options.sort_by{|de| de['delivery_type']['name'].to_i})
   end
 
   # POST /sellers/:seller_id/delivery_options
--- a/app/controllers/delivery_options_controller.rb
+++ b/app/controllers/delivery_options_controller.rb
@@ -1,0 +1,55 @@
+class DeliveryOptionsController < ApplicationController
+  before_action :set_seller
+  before_action :set_seller_delivery_options, only: [:update, :destroy]
+
+  # GET /sellers/:seller_id/delivery_options
+  def index
+    json_response(@seller.delivery_options)
+  end
+
+  # POST /sellers/:seller_id/menu_items
+  def create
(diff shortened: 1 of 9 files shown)


spec/requests/delivery_options_request_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+      {
+        invalid: 'quick brown fox',
+      }
+    end
+
+    context 'when request attributes are valid' do
+      before { post "/sellers/#{seller_id}/delivery_options", params: valid_attributes, as: :json }
+
+      it 'creates a delivery_option' do
+        actual_json = json.except('id')
```
justintmckibben, line 57: nit: could we also check 

```id = json['id']
expected_delivery_option = DeliveryOption.find(json['id'])
expect(expected_delivery_option).to exist
```

justintmckibben:
For DeliveryType you don't *need* to if you don't mind creating them in the prod DB and seeding them

app/controllers/delivery_options_controller.rb
```diff
@@ -0,0 +1,55 @@
+class DeliveryOptionsController < ApplicationController
+  before_action :set_seller
+  before_action :set_seller_delivery_options, only: [:update, :destroy]
+
+  # GET /sellers/:seller_id/delivery_options
+  def index
+    json_response(@seller.delivery_options)
```
justintmckibben, line 7: If you don't expose delivery_type on a different endpoint, you'd have to return it from this endpoint otherwise the frontend won't have any way to get the name and icon_url

juliexxia: Would it be better to expose it on a different controller or expose it here? I imagine here? What would that look like? I kind of assumed `@seller.delivery_options` would also expose the delivery type but I guess that's not true?

app/controllers/delivery_options_controller.rb
```diff
@@ -0,0 +1,62 @@
+class DeliveryOptionsController < ApplicationController
+  before_action :set_seller
+  before_action :set_seller_delivery_options, only: [:update, :destroy]
+
+  # GET /sellers/:seller_id/delivery_options
+  def index
+    delivery_options = @seller.delivery_options.map do |delivery_option|
+      delivery_type = delivery_option.delivery_type
+      json = delivery_option.as_json
+      json['delivery_type'] = delivery_type.as_json
```
justintmckibben, line 10: Mind testing this in a unit test?

juliexxia: Yeah I was having trouble finding an example of a unit test of a has_* association - any advice?

spec/requests/delivery_options_request_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+  describe 'GET /sellers/:seller_id/delivery_options' do
+    before { get "/sellers/#{seller_id}/delivery_options" }
+
+    context 'when seller exists' do
+      it 'returns status code 200' do
+        expect(response).to have_http_status(200)
+      end
+
+      it 'returns all seller delivery_options' do
+        expect(json.size).to eq(20)
```
justintmckibben, line 22: expect(json).to eq(expected_json) # where expected json includes the conversion of delivery_type into a json too

app/controllers/delivery_options_controller.rb
justintmckibben: https://github.com/sendchinatownlove/ruby/pull/178/files#r460165708

spec/requests/delivery_options_request_spec.rb
juliexxia: getting lost in the sauce here: the delivery_option delivery_type relationship is `delivery_option has_one delivery_type` but `delivery_type **does not** belong_to delivery_option`

The delivery_options are being created by factorybot and looking through all our factories and tests I'm not sure how to set up a `delivery_option` that is linked to a `delivery_type` when `delivery_type`s don't belong to `delivery_option`s, if that makes sense.

justintmckibben: I think the relationship might be wrong. It should be that a delivery_option belongs to delivery_type

juliexxia: Hm, the current relationship is

a `delivery_option` `has_one` `delivery_type`
a `seller` `has_many` `delivery_option`s

with schemas

`delivery_option`
- `phone number`
- `url` 

delivery_type
- `name` e.g. "gruhub"

Was your thought that it would be delivery_type has_many delivery_options? I guess that would work too. I feel like how it is now makes slightly more sense?

juliexxia: Err, I guess each `delivery_option` could `belong_to` a `delivery_type` without the reciprocal `delivery_type` `has_many` `delivery_option`s

app/controllers/delivery_options_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+      json
+    end
+
+    json_response(delivery_options.sort_by{|de| de['delivery_type']['name'].to_i})
```
justintmckibben, line 14: There is a way to do this in the query, but maybe you can do that as a follow up to unblock Olivia? Just create a separate Trello task. But here we are doing a N log N sort for every Seller. Not terrible but worth filing

justintmckibben:
Oh do you mind running `rubocop -a .`?

app/controllers/delivery_options_controller.rb
juliexxia: https://trello.com/c/21DtS7WI

juliexxia:
Promise: will run rubocop on everything after submit (:

juliexxia:
As promised: https://github.com/sendchinatownlove/ruby/pull/183

PR #179, opened 2020-07-20
Rake task to send giftcards links to all customers

Jeremy-Xue:
Example screenshot for email sent from the seeded db (formatting is a little weird since I sent a bunch of emails beforehand to test but it shouldn't be purple/spaced in the actual email)
Screen Shot 2020-07-20 at 4 34 18 PM

Looking for suggestions on how best to word/format the email, as well as if we want to add an argument to only send emails to those who haven't already gotten the link (before the PR that changed the emails to link to the giftcard). Also, unsure how this interacts with GAM meals (do those count as gift card purchases?)

### DIFF
--- a/app/services/email_manager/sender.rb
+++ b/app/services/email_manager/sender.rb
@@ -18,12 +18,31 @@ module EmailManager
                       subject: 'Receipt from Send Chinatown Love',
                       html: html
 
-      # Send to Receipts Eng so that they know what their receipts in prod looks
-      # like
+      # Send to Receipts Eng so that they know what their receipts in prod looks like
       RestClient.post api_url,
                       from: 'receipts@sendchinatownlove.com',
                       to: 'receipts@sendchinatownlove.com',
                       subject: 'Receipt from Send Chinatown Love',
+                      html: html
+    end
+
+    def self.send_receipt_with_subject(to:, html:, subject:)
+      api_key = ENV['MAILGUN_API_KEY']
+      # rubocop:disable Layout/LineLength
+      api_url = "https://api:#{api_key}@api.mailgun.net/v2/m.sendchinatownlove.com/messages"
+      # rubocop:enable Layout/LineLength
+
+      RestClient.post api_url,
+                      from: 'receipts@sendchinatownlove.com',
+                      to: to,
+                      subject: subject,
+                      html: html
+
+      # Send to Receipts Eng so that they know what their receipts in prod looks like
+      RestClient.post api_url,
+                      from: 'receipts@sendchinatownlove.com',
+                      to: 'receipts@sendchinatownlove.com',
+                      subject: subject,
                       html: html
     end
   end
--- a/lib/tasks/email_giftcards.rake
+++ b/lib/tasks/email_giftcards.rake
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-namespace :email_giftcards do
+namespace :emailer do
   desc 'Email customers the gift cards that they ordered'
-  task send_emails: :environment do
+  task vouchers_to_customers: :environment do
     desc 'send emails'
 
     unique_recipients = GiftCardDetail
--- a/lib/tasks/email_giftcards.rake
+++ b/lib/tasks/email_giftcards.rake
@@ -17,7 +17,7 @@ namespace :email_giftcards do
               "  <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />" \
               '</head>' \
               '<body>' \
-              '<p> Thank you for supporting local chinatown businesses! As our merchants open up, you will be able to redeem the giftcards you have purchased below.</p>'
+              '<p> Thank you for supporting local Chinatown businesses!</p>'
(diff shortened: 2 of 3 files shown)


justintmckibben:
cc @nanxiy do you have opinions on the language that would be used here?

lib/tasks/email_giftcards.rake
```diff
@@ -0,0 +1,36 @@
...
+                        .distinct.pluck(:recipient_id)
+
+    unique_recipients.each do |id|
+      html = '<!DOCTYPE html>' \
+              '<html>' \
+              '<head>' \
+              "  <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />" \
+              '</head>' \
+              '<body>' \
+              '<p> Thank you for supporting local chinatown businesses! As our merchants open up, you will be able to redeem the giftcards you have purchased below.</p>'
```
justintmckibben, line 20: Chinatown is a properly noun

justintmckibben: I'd consider making it simpler and just removing this sentence `As our merchants open up, you will be able to redeem the giftcards you have purchased below.`

justintmckibben: Thank you for supporting local Chinatown businesses! Your gift card to <merchant name> can be found here: <link to gift card>

Current balance: $XX.XX

justintmckibben: But I'd just swing it by @nanxiy and see what she thinks

jacoblb64: > Chinatown is a properly noun

I like how you called out his grammar with your own grammatical mistake 😂. But in all seriousness good catch

lib/tasks/email_giftcards.rake
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+namespace :email_giftcards do
```
jacoblb64, line 3: Optional, but if you want you can put this in the `emailer.rake` namespace where the other task was. That way the tasks can be called something like `rake emailer:vouchers_to_merchants` or `emailer:vouchers_to_customers` or maybe even just rename it `email` to make it even more succinct. I trust you to choose what's best!

lib/tasks/email_giftcards.rake
```diff
@@ -0,0 +1,36 @@
...
+              "  <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />" \
+              '</head>' \
+              '<body>' \
+              '<p> Thank you for supporting local chinatown businesses! As our merchants open up, you will be able to redeem the giftcards you have purchased below.</p>'
+      recipient = Contact.find_by(id: id)
+      gift_cards = GiftCardDetail.where(recipient_id: id).to_a
+      html += '<style> table {border: 1px solid black}</style>'\
+            '<table>'
+      gift_cards.each do |gc|
+        gift_card_url = 'merchant.sendchinatownlove.com/voucher/' + gc.gift_card_id
```
jacoblb64, line 26: @justintmckibben should this be `seller_gift_card_id`? This still confuses me about which code gets used where tbh

lib/tasks/email_giftcards.rake
```diff
@@ -0,0 +1,36 @@
...
+        amount_string = EmailManager::Sender.format_amount(amount: gc.amount)
+        item = Item.find_by(id: gc.item_id)
+        merchant_name = Seller.find_by(id: item.seller_id).name
+        html += '<tr><p>View your <b>$' + amount_string + '</b> gift card for <b>' + merchant_name + '<a href="' + gift_card_url + '"> here</a></b></p></tr>'
+      end
+      html += '</table>'
+      EmailManager::Sender.send_receipt(to: recipient.email, html: html)
```
jacoblb64, line 33: Optional, but might be helpful to log some output so we can track what emails get sent to whom. I did some basic logging in the task in my last PR if you want to take a look.

Jeremy-Xue:
Ack I just realized I never ended up merging this -- I was working on adding a bit to only send the gift card emails to people who bought them before the date we added the gift card url to emails, also I think I need to filter out GAM (?)

app/services/email_manager/sender.rb
```diff
@@ -18,13 +18,32 @@ def self.send_receipt(to:, html:)
...
-      # like
+      # Send to Receipts Eng so that they know what their receipts in prod looks like
       RestClient.post api_url,
                       from: 'receipts@sendchinatownlove.com',
                       to: 'receipts@sendchinatownlove.com',
                       subject: 'Receipt from Send Chinatown Love',
                       html: html
     end
+
+    def self.send_receipt_with_subject(to:, html:, subject:)
```
Jeremy-Xue, line 29: I'm wondering if this is adding too much redundancy but I think this was the easiest way to send an email with a custom subject

Jeremy-Xue: @justintmckibben was using this to send the gift card email with a custom subject, should we use this for the passport/food crawl redemption email?

PR #180, opened 2020-07-23
Create migration and model for campaigns

rjvani:


db/migrate/20200723225140_create_campaigns.rb
```diff
@@ -0,0 +1,16 @@
+class CreateCampaigns < ActiveRecord::Migration[6.0]
+  def change
+    create_table :campaigns do |t|
+      t.boolean :active, default: false
+      t.boolean :is_valid, default: false
```
justintmckibben, line 5: Could we make this valid or make active is_active?

db/migrate/20200723225140_create_campaigns.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.datetime :end_date, null: false
+      t.string :description, null: false
```
justintmckibben, line 7: I think we can make this nullable

db/migrate/20200723225140_create_campaigns.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.string :gallery_image_urls, default: [], null: false, array: true
```
justintmckibben, line 8: This could also be nullable. Maybe we can put a default gallery image in the frontend

db/migrate/20200723225140_create_campaigns.rb
rjvani: I had an issue earlier because I think `ActiveRecord` has a `valid` method. I installed [`safe_attributes`](https://github.com/bjones/safe_attributes) and updated to `valid`.

PR #181, opened 2020-07-24
Create campaign controller

rjvani:
Add GET, POST, and PUT endpoints for `Campaign`s

### DIFF
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+class CampaignsController < ApplicationController
+  before_action :set_campaign, only: %i[show update]
+
+  # GET /campaigns
+  def index
+    json_response(Campaign.all)
+  end
+
+  # GET /campaigns/:id
+  def show
+    json_response(@campaign)
+  end
+  
+  # POST /campaigns
+  def create
+    campaign = Campaign.create!(create_params)
+    json_response(campaign, :created)
+  end
+
+  # PUT /campaigns/:id
+  def update
+    @campaign.update(update_params)
+    json_response(@campaign)
+  end
+
+  private
+
+  def create_params
+    params.require(:end_date)
+    params.require(:location_id)
+    params.require(:seller_id)
+
+    ret = params.permit(
+      :active,
+      :description,
+      :end_date,
+      :valid,
+      gallery_image_urls: []
+    )
+
+    set_location
+    set_seller
+    ret[:location_id] = @location.id
+    ret[:seller_id] = @seller.id
+
+    ret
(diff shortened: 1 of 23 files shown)


spec/requests/campaigns_request_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      end
+
+      it 'Returns 200' do
+        expect(response).to have_http_status(200)
+      end
+    end
+  end
+
+  context 'POST /campaigns' do
+    context 'With invalid parameters' do
```
justintmckibben, line 51: Should we test invalid seller_id and location_id?

spec/factories/campaign.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :campaign do
+    active { true }
+    valid { true }
+    end_date { DateTime.now }
```
justintmckibben, line 7: Faker has a way to generate a random date in the future: `Faker::Time.forward(days: 30)   # any time up to 30 days into the future`

spec/requests/campaigns_request_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+          '/campaigns',
+          params: {
+            end_date: Date.tomorrow,
+            location_id: @location.id,
+            seller_id: @seller.seller_id,
+          },
+          as: :json,
+        )
+      end
+
```
justintmckibben, line 71: could you add a test that Campaigns.all.size is 1? and maybe has the attributes you passed in

app/controllers/campaigns_controller.rb
```diff
@@ -0,0 +1,72 @@
+# frozen_string_literal: true
+
+class CampaignsController < ApplicationController
+  before_action :set_campaign, only: %i[show update]
+
+  # GET /campaigns
+  def index
+    json_response(Campaign.all)
```
justintmckibben, line 8: We should Order by end_date where the one that is ending soonest is the first

spec/requests/campaigns_request_spec.rb
rjvani: Good call, added

spec/requests/campaigns_request_spec.rb
rjvani: Added a test to look up the new record and verify attributes

app/controllers/campaigns_controller.rb
rjvani: Added the order, but would we need to filter on `active` also?

rjvani: Added a query parameter for this endpoint

PR #182, opened 2020-07-27
Add distributor and migrate to use campaign models

justintmckibben:


justintmckibben:
Needs tests but here is a start of migrating to the new model of distributor

db/migrate/20200726210802_create_distributors.rb
```diff
@@ -0,0 +1,8 @@
+class CreateDistributors < ActiveRecord::Migration[6.0]
+  def change
+    create_table :distributors do |t|
+      t.string :website_url
```
danthaman44, line 4: do we want to add name or some other kind of unique identifier?

app/controllers/charges_controller.rb
```diff
@@ -15,16 +15,29 @@ def create
...
+      )
+    elsif charge_params[:campaign_id].present?
+      Campaign.find_by(campaign_id: campaign_id)
+    end
 
-    validate(seller_id: seller_id, line_items: line_items, is_distribution: is_distribution)
+    validate(
+      seller_id: seller_id,
+      line_items: line_items,
+      gift_a_meal: campaign.present?
```
danthaman44, line 33: love how we're changing `is_distribution` to `gift_a_meal`, way more understandable

db/migrate/20200726210802_create_distributors.rb
justintmckibben: The name lives on the Contact

app/controllers/charges_controller.rb
```diff
@@ -72,14 +69,16 @@ def charge_params
       :seller_id,
       :idempotency_key,
       :is_subscribed,
+      :campaign_id,
+      # TODO(justintmckibben): Deprecate this boolean in favor of campaign_id
```
justintmckibben, line 73: FYI @oliviasztanga @danthaman44 for GAM purchases, you have to pass the campaign_id instead of the boolean is_distribution on the frontend. is_distribution will still work, but it will just find the first Campaign that it finds

app/helpers/sellers_helper.rb
```diff
@@ -3,12 +3,13 @@
 module SellersHelper
   def self.generate_seller_json(seller:)
     locations = seller.locations
-    distributor = seller.distributor
+    distributor = get_deprecated_distributor(seller: seller)
```
justintmckibben, line 6: Also FYI @oliviasztanga @danthaman44 the distributor on the /sellers/ endpoint is being deprecated in favor of the /campaigns/ endpoint which should include distributor

db/migrate/20200726220247_migrate_distributor_to_new_distributor.rb
```diff
@@ -0,0 +1,26 @@
...
+      next unless contact.present?
+
+      # NB(justintmckibben): Remember to update the end_date of past campaigns
+      # to the real end_date
+      campaign = Campaign.create valid: true, end_date: Time.now
+      campaign.seller = seller
+      campaign.distributor = Distributor.create contact: contact
+    end
+
+    remove_column :contacts, :seller_id
```
rjvani, line 24: Maybe can push this off to a future migration depending on how deploys are done. I'll follow up in Slack.

app/controllers/charges_controller.rb
```diff
@@ -109,23 +108,44 @@ def validate(seller_id:, line_items:, is_distribution:)
...
+                      seller_id: @seller.id,
+                      active: true,
+                      valid: true
+                    )
+                    unless campaign
+                      raise ActiveRecord::RecordNotFound, "Passed in is_distribution with no active campaign running for seller_id=#{seller_id}"
+                    end
+
+                    campaign
+                  elsif charge_params[:campaign_id].present?
```
justintmckibben, line 124: FYI this is an untested codepath. I'll follow up with a test for this + will add a seed file

PR #184, opened 2020-08-03
Update Seller Gift Card endpoint to filter out GAM meals

Jeremy-Xue:
Works on the assumption that all GAM gift cards will have single_use: true, and that no non-GAM gift cards will have single_use: true

Updated spec tests to include a single_use giftcard, and seedfile to include a single_use giftcards

### DIFF
--- a/app/controllers/seller_gift_cards_controller.rb
+++ b/app/controllers/seller_gift_cards_controller.rb
@@ -39,8 +39,8 @@ class SellerGiftCardsController < ApplicationController
   def set_seller
     @seller = Seller.find_by!(seller_id: params[:seller_id])
 
-    #if @seller.gift_cards_access_token != params[:id]
-    #  raise ActiveRecord::RecordNotFound
-    #end
+    if @seller.gift_cards_access_token != params[:id]
+      raise ActiveRecord::RecordNotFound
+    end
   end
 end
--- a/app/controllers/seller_gift_cards_controller.rb
+++ b/app/controllers/seller_gift_cards_controller.rb
@@ -13,8 +13,6 @@ class SellerGiftCardsController < ApplicationController
               :email,
               :created_at,
               :expiration,
-            ).where(
-              single_use: false
             )
             .joins(:item, :recipient)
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -94,26 +94,36 @@
     email: 'testytesterson@gmail.com',
     item_type: Item.gift_card,
     refunded: false,
+    single_use: false,
```
Jeremy-Xue, line 97: I think the default value for this is false, so I can take it out if we want

app/controllers/seller_gift_cards_controller.rb
```diff
@@ -12,7 +12,9 @@ def show
               :name,
               :email,
               :created_at,
-              :expiration
+              :expiration,
+            ).where(
+              single_use: false
```
justintmckibben, line 17: I think we want to allow for the frontend to specify whether or not it wants gift_a_meal cards or not. Like a query parameter. @Yong-L has an example of ordering for our Sellers — so something akin to that except a filter instead of an order

Jeremy-Xue: gotcha! can make the change

Jeremy-Xue: @Yong-L I was wondering if creating a validation file was necessary, I tried it out by adding an if statement to check for parameters to append the .where clause to the query, which worked when I tested it out in postman, but I was wondering if I was missing something by forgoing a separate query generator in the validate folder.

db/seeds.rb
jacoblb64: I think it's nice to be explicit for everyone reading the file. What do you think @justintmckibben ?

Jeremy-Xue:
> https://github.com/sendchinatownlove/ruby/pull/193/files#diff-9c9986a1c69e76075d0e3ce491fbd434R13

Oo good catch! I was actually wondering about this earlier, haha. We can maybe hold off on merging this then and I can take a stab at changing it to check for the campaign association instead

PR #187, opened 2020-08-06
Add target amount to campaign

justintmckibben:


db/seeds.rb
```diff
@@ -141,3 +141,13 @@
   fee_attributes[:seller_id] = seller.id
   Fee.create!(fee_attributes)
 end
+
+seller = Seller.find_by(seller_id: 'shunfa-bakery')
```
danthaman44, line 145: This is huge, no more manually creating campaigns in my local DB :)

PR #188, opened 2020-08-07
Adding a new "original amounts" query and using it for the seller progress bar

jacoblb64:
- When doing the `distinct on` query, order by ascending (oldest) amounts instead of descending (newest).
- Keeping both queries as we still use the `latest` for the merchant dashboard and both are useful
- Switching seller endpoint to calculate based on original, not latest amounts
- Updated unit tests with the new expected amounts
- running `rails db:migrate` wanted to move a schema row, so I let it


justintmckibben:
thanks for picking this up so quickly!

PR #190, opened 2020-08-07
Add name field to distributor

danthaman44:
Add the organization name to the distributor model. This field is required to render the GAM campaign title of the format "Merchant name x Distributor name"
1. The distributor contact name is not reliable since it's often the name of a person and not the org.
2. Reduces the # of server calls required from the front-end. Only one call to the server to fetch the distributor is needed, as opposed to two calls to fetch the distributor and the contact.

Test plan
```
danthaman-mac:ruby danthaman$ heroku local:run bundle exec rspec spec/requests/distributors_request_spec.rb
........

Finished in 0.40248 seconds (files took 2.05 seconds to load)
8 examples, 0 failures
```

Screen Shot 2020-08-06 at 11 29 44 PM

### DIFF
--- a/app/controllers/distributors_controller.rb
+++ b/app/controllers/distributors_controller.rb
@@ -18,7 +18,8 @@ class DistributorsController < ApplicationController
     params.permit(
       :website_url,
       :image_url,
-      :contact_id
+      :contact_id,
+      :name
     )
   end
 
--- a/app/models/distributor.rb
+++ b/app/models/distributor.rb
@@ -6,6 +6,7 @@
 #
 #  id          :bigint           not null, primary key
 #  image_url   :string
+#  name        :string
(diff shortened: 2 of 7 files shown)


app/controllers/distributors_controller.rb
```diff
@@ -18,7 +18,8 @@ def create_params
     params.permit(
       :website_url,
       :image_url,
-      :contact_id
+      :contact_id,
+      :name
     )
   end
```
justintmckibben, line 24: Ahhh can you actually just have distributor return distributor.contact.name?

danthaman44: Are you sure we want to tie these values together? I thought the contact could just be a random person's name (ie. Joe from Apex)

PR #193, opened 2020-08-08
Add campaign to existing items

justintmckibben:


db/migrate/20200808054745_add_campaign_to_items.rb
```diff
@@ -0,0 +1,19 @@
...
+  def change
+    Seller.all.each do |seller|
+      # Currently there is max one campaign per Seller
+      campaign = Campaign.find_by(seller_id: seller.id)
+      # We want to modify all Items including the refunded ones to be associated
+      # to the correct campaign
+      item = Item.joins(:gift_card_detail).where(items: { seller_id: seller.id, single_use: true })
+      item.update(campaign: campaign)
+      item.payment_intent.update(campaign: campaign)
+      # NB(justintmckibben): After we run this migration, single_use isn't the
```
jacoblb64, line 13: Are we removing `single_use` then, or just not setting it going forward? I'm curious to hear more about how to keep this clean

justintmckibben: We are going to set it, it just has a new meaning. We don't do this now, but this will allow for us to sell single use gift cards vs it being reserved for only GAM campaigns.

Jeremy-Xue: should we merge the existing filter and edit it later once we do start using single use giftcards?

justintmckibben: Right now single_use still implies GAM, but that might not always be the case going forward. I can see a world where for future Campaigns that don't require gift cards to be created, we create donations items instead of gift cards, and know that it's a GAM donation because of the presence of a campaign. However for now, yeah all gif cards that are single_use are GAM.

For your other PR, I think since we're operating under that assumption it's fine to go ahead and merge. Then we can add to the PR and also allow for us to filter out GAM by campaign. That'll mean that the old filter just is truly filtering out single_use gift cards, which also might be a useful filter to the merchants.

PR #196, opened 2020-08-08
Fix items migration

justintmckibben:
The where was returning an array and we need to
apply the migration to each element in the query

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -94,26 +94,36 @@ end
     email: 'testytesterson@gmail.com',
     item_type: Item.gift_card,
     refunded: false,
-    amounts: [10_000, 8000]
+    amounts: [10_000, 8000],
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -94,26 +94,36 @@
...
   {
     email: 'testytesterson3@gmail.com',
     item_type: Item.gift_card,
     refunded: false,
-    amounts: [5000]
+    amounts: [5000],
+    single_use: false
+  },
+  {
+    email: 'testytesterson3@gmail.com',
```
jacoblb64, line 115: nit: should we make this testy testerson 4? 😜

justintmckibben: LOL yes

jacoblb64: 😎

PR #197, opened 2020-08-10
Create Tickets for SCL Crawl

jacoblb64:
Honestly, this is my first time creating a rails table from scratch, let me know if there's anything I did wrong or that we do differently.

Work completed so far
- Ticket class
- Migration
- Schema update
- Sample seed data (used some faked names, and tried to get all the optional possibilities)
- Basic testing and factories
- Running rubocop on everything (mostly tweaked @Jeremy-Xue's recent stuff)

[Relevant RFC](https://docs.google.com/document/d/1HqGgOQoJtvLgstOISq1Ifk6c3N0VCrmmSzqH7LjRZHE/edit?usp=sharing)

**Todo**: create endpoint(s)

Ticket table:
Screen Shot 2020-08-09 at 10 06 58 PM

And new contacts created with seed data:
Screen Shot 2020-08-09 at 10 07 08 PM


db/migrate/20200809160820_create_tickets.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class CreateTickets < ActiveRecord::Migration[6.0]
+  def change
+    create_table :tickets do |t|
+      t.references :contact, null: true, foreign_key: true
+      t.string :ticket_id
+      t.references :participating_seller, references: :sellers, null: false, foreign_key: { to_table: :sellers}
+      t.references :sponsor_seller, references: :sellers, null: true, foreign_key: { to_table: :sellers}
```
Jeremy-Xue, line 9: i think sponsor seller isn't going to be related to sellers, but participating sellers can be but is nullable!

justintmckibben: participating sellers will reference seller directly, so both should just reference the corresponding participating/sponsor seller. This should have no reference to seller directly

justintmckibben: also i think this should be null: false since you can only redeem a ticket at a sponsor seller

db/migrate/20200809160820_create_tickets.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class CreateTickets < ActiveRecord::Migration[6.0]
+  def change
+    create_table :tickets do |t|
+      t.references :contact, null: true, foreign_key: true
```
justintmckibben, line 6: all tickets should be owned by a contact aka someone should be redeeming it so null: false

db/migrate/20200809160820_create_tickets.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.string :ticket_id
```
justintmckibben, line 7: also null: false. Not sure if unique: true is a thing, but it needs a unique enforcement

spec/models/ticket_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+#  fk_rails_...  (participating_seller_id => sellers.id)
+#  fk_rails_...  (sponsor_seller_id => sellers.id)
+#
+require 'rails_helper'
+
+RSpec.describe Ticket, type: :model do
+  # Association tests
+  it { should belong_to(:contact) }
+  it { should belong_to(:participating_seller) }
+end
```
justintmckibben, line 35: there are tests to validate uniqueness/presence of

db/migrate/20200809160820_create_tickets.rb
jacoblb64: Updated with the right references, apologies for the earlier confusion! @justintmckibben, could you explain your last comment? Isn't it the opposite? Every ticket needs a ParticipatingSeller (ticket is associated with a merchant that gives it out) but starts null with a SponsorSeller (reward)?

db/migrate/20200809160820_create_tickets.rb
jacoblb64: 👍

db/migrate/20200809160820_create_tickets.rb
jacoblb64: Doesn't seem to be a way to make a non-index column here unique, but it's enforced in the ticket model

spec/models/ticket_spec.rb
jacoblb64: Sweet! Took inspiration from `gift_card_detail` and tried to match

justintmckibben:
Great tests too!

db/migrate/20200809160820_create_tickets.rb
jacoblb64: I think we figured it out :)

PR #199, opened 2020-08-11
[WIP] Passport SponsorSeller/ParticipatingSeller model

Jeremy-Xue:
WIP model code

Based on RFC: https://docs.google.com/document/d/1HqGgOQoJtvLgstOISq1Ifk6c3N0VCrmmSzqH7LjRZHE

### DIFF
--- a/app/models/participating_seller.rb
+++ b/app/models/participating_seller.rb
@@ -11,5 +11,5 @@
 #  seller_id      :bigint
 #
 class ParticipatingSeller < ApplicationRecord
-    validates_presence_of :name, :tickets_secret
+    validates_presence_of :name
 end
--- a/app/models/participating_seller.rb
+++ b/app/models/participating_seller.rb
@@ -1,0 +1,15 @@
+# == Schema Information
+#
+# Table name: participating_sellers
+#
+#  id             :bigint           not null, primary key
+#  name           :string
+#  stamp_url      :string
+#  tickets_secret :string
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+#  seller_id      :bigint
+#
+class ParticipatingSeller < ApplicationRecord
+    validates_presence_of :name, :tickets_secret
+end
--- a/app/models/sponsor_seller.rb
+++ b/app/models/sponsor_seller.rb
@@ -6,11 +6,11 @@
 #  logo_url    :string
 #  name        :string
 #  reward      :string
-#  reward_cost :integer
+#  reward_cost :integer          default(3), not null
 #  created_at  :datetime         not null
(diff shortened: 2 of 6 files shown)


db/migrate/20200811003516_create_participating_sellers.rb
```diff
@@ -0,0 +1,12 @@
+class CreateParticipatingSellers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :participating_sellers do |t|
+      t.string :name
+      t.bigint :seller_id, null: true
```
Jeremy-Xue, line 5: actually I just realized, I think null defaults to true, should the other options (name stamp_url etc) be set to null:false instead?

jacoblb64: sure, makes sense to me, whatever's default can be default, no need to manually specify

app/models/sponsor_seller.rb
```diff
@@ -0,0 +1,16 @@
...
+#  logo_url    :string
+#  name        :string
+#  reward      :string
+#  reward_cost :integer
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#  location_id :bigint
+#
+class SponsorSeller < ApplicationRecord
+    validates_presence_of :logo_url, :name, :reward, :reward_cost
```
justintmckibben, line 15: I think :logo_url, :name, :reward you don't need to validate the presence of. Definitely not :logo_url and :reward — but I can see the argument of you wanting to validate name is there

db/migrate/20200811003054_create_sponsor_sellers.rb
```diff
@@ -0,0 +1,13 @@
+class CreateSponsorSellers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :sponsor_sellers do |t|
+      t.string :name
+      t.bigint :location_id
+      t.string :logo_url
+      t.string :reward
+      t.integer :reward_cost
```
justintmckibben, line 8: I'd default this to three, and make it nullable: false

app/models/participating_seller.rb
```diff
@@ -0,0 +1,15 @@
...
+#  id             :bigint           not null, primary key
+#  name           :string
+#  stamp_url      :string
+#  tickets_secret :string
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+#  seller_id      :bigint
+#
+class ParticipatingSeller < ApplicationRecord
+    validates_presence_of :name, :tickets_secret
```
justintmckibben, line 14: tickets_secret you don't need to validate the presence of. Matter of fact we want for it to be nullable

PR #200, opened 2020-08-13
Add new fields to Contact for Passport

justintmckibben:
Added instagram and access key
Made email unique
Added a controller to give ability to create
new contact so that frontend can do so when
registering a ticket


justintmckibben:
Added client engineers so y'all can take a look at the API

PR #201, opened 2020-08-14
Add neighborhood field to location model

danthaman44:
The new GAM page shows the merchant's neighborhood ("Chinatown", "Sunset Park") in addition to their city/borough. Adding this field
Screen Shot 2020-08-14 at 12 00 16 AM

Test plan:
Screen Shot 2020-08-13 at 11 56 15 PM

ruby/spec/requests/locations_spec.rb passes

### DIFF
--- a/db/migrate/20200814034110_add_neighborhood_to_locations.rb
+++ b/db/migrate/20200814034110_add_neighborhood_to_locations.rb
@@ -1,0 +1,5 @@
+class AddNeighborhoodToLocations < ActiveRecord::Migration[6.0]
+  def change
+    add_column :locations, :neighborhood, :string
+  end
+end
--- a/app/controllers/locations_controller.rb
+++ b/app/controllers/locations_controller.rb
@@ -37,7 +37,7 @@ class LocationsController < ApplicationController
   end
 
   def update_location_params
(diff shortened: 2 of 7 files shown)


db/migrate/20200814034110_add_neighborhood_to_locations.rb
```diff
@@ -0,0 +1,5 @@
+class AddNeighborhoodToLocations < ActiveRecord::Migration[6.0]
+  def change
+    add_column :locations, :neighborhood, :string
```
justintmckibben, line 3: Should we also add borough while we're at it? I know that we sort of sub city for that right now but it's not quite accurate

justintmckibben: not a blocker

justintmckibben: I guess we can also just use city as borough like we're already doing

danthaman44: Yes good call. I did notice that we're using the "City" field to represent borough, which is not scalable lol. Will have a followup PR out for that soon.

PR #210, opened 2020-08-17
Add Participating Sellers API

Jeremy-Xue:


spec/requests/participating_sellers_request_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+      )
+    end
+
+    context 'with valid attrs' do
+      let(:attrs) do
+        {
+          seller_id: seller.id,
+          name: '46 Mott',
+          stamp_url: 'sendchinatownlove.com/test_stamp_url',
+          tickets_secret: 'supersecret',
```
justintmckibben, line 69: we shouldn't let them set their secret from the API

app/controllers/participating_sellers_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  private
+
+  def participating_seller_params
+    params.permit(
+      :name,
+      :seller_id,
+      :stamp_url,
+      :tickets_secret
```
justintmckibben, line 31: This shouldn't be settable from the public API

app/controllers/participating_sellers_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  # POST /participating_sellers
+  def create
+    json_response(ParticipatingSeller.create!(participating_seller_params), :created)
+  end
+
+  # GET /participating_sellers/:id
+  def show
+    json_response(@participating_seller)
```
justintmckibben, line 15: We also shouldn't return the secret either — it's like returning their password

app/controllers/participating_sellers_controller.rb
```diff
@@ -0,0 +1,38 @@
+class ParticipatingSellersController < ApplicationController
+  before_action :set_participating_seller, only: %i[show update]
+
+  def index
+    json_response(ParticipatingSeller.all)
+  end
+
+  # POST /participating_sellers
+  def create
```
justintmckibben, line 9: We should generate and set the secret right here. You can generate a SHA the same way we do for our gift cards here

Jeremy-Xue: I added it to the model like for sellers gift cards, does that work?

justintmckibben: yeah

PR #211, opened 2020-08-17
Tickets endpoints

jacoblb64:
- `index` and `show` endpoints to get list of or a specific ticket
- Can create any number of tickets, need to pass a participating seller (Note: might refactor this to be on participating seller instead)
- Can update a ticket, but only with contact info
- Migration and code change to make `contact` optional on a ticket, as it was mistakenly set to non-nullable / required
- Test coverage for all the new endpoints, including valid and multiple invalid cases

Also running `rubocop -a`, affecting a lot of new files


db/migrate/20200817030307_make_contact_nullable_on_tickets.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class MakeContactNullableOnTickets < ActiveRecord::Migration[6.0]
+  def change
+    change_column :tickets, :contact_id, :bigint, null: true
```
justintmckibben, line 5: oops

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+
+class TicketsController < ApplicationController
+  before_action :set_ticket, only: %i[show update]
+  before_action :set_participating_seller, only: %i[create]
+
+  # GET /tickets
+  def index
+    @tickets = Ticket.all
+
+    render json: @tickets
```
justintmckibben, line 11: Could you use json_response instead? It's defined in Response to be exactly this

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  # GET /tickets/1
+  def show
```
justintmckibben, line 15: Yeah this should be fine

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+    render json: @ticket
+  end
+
+  # POST /tickets
+  def create
+    puts(params[:number_of_tickets])
+    savedCorrectly = true
+    createdTickets = []
+
+    (1..params[:number_of_tickets]).each do |_i|
```
justintmckibben, line 25: Is this really inclusive?

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      attributes = {}
+      attributes[:ticket_id] = SecureRandom.alphanumeric(5).upcase.insert(4, '-')
```
justintmckibben, line 27: Will they have to enter the dash when they enter it into the passport app? Should we make them? cc @nanxiy

app/controllers/tickets_controller.rb
justintmckibben, line 25: We should wrap this in a transaction and not create any tickets if one of them fail. Also could we add a test that tests that if it fails it doesn't create any tickets?

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      attributes[:participating_seller] = @participating_seller
+
+      @ticket = Ticket.new(attributes)
+      savedCorrectly &&= @ticket.save
```
justintmckibben, line 31: Does this work like you think it does? I think it saves upon create.

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      createdTickets << @ticket
+    end
+
+    if savedCorrectly
+      render json: createdTickets, status: :created
+    else
+      render 'Error', status: :unprocessable_entity
```
justintmckibben, line 38: Does this really get hit? It should automatically fail and get caught in the exception_handler

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+    puts(params[:number_of_tickets])
+    savedCorrectly = true
+    createdTickets = []
+
+    (1..params[:number_of_tickets]).each do |_i|
+      attributes = {}
+      attributes[:ticket_id] = SecureRandom.alphanumeric(5).upcase.insert(4, '-')
+      attributes[:participating_seller] = @participating_seller
+
+      @ticket = Ticket.new(attributes)
```
justintmckibben, line 30: I think there is a way you can make this fail if it doesn't create with a bang!

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      render 'Error', status: :unprocessable_entity
+    end
+  end
+
+  # PATCH/PUT /tickets/1
+  def update
+    if @ticket.update(ticket_params)
+      render json: @ticket
+    else
+      render json: @ticket.errors, status: :unprocessable_entity
```
justintmckibben, line 47: Don't think you need this catch with an else

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  # GET /tickets/1
+  def show
+    render json: @ticket
+  end
+
+  # POST /tickets
+  def create
+    puts(params[:number_of_tickets])
```
justintmckibben, line 21: note to remember to delete

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  private
+
+  # Use callbacks to share common setup or constraints between actions.
+  def set_ticket
+    @ticket = Ticket.find(params[:id])
+  end
+
+  def set_participating_seller
+    @participating_seller = ParticipatingSeller.find_by!(id: params[:participating_seller_id])
```
justintmckibben, line 59: ParticipatingSeller.find(params[:participating_seller_id])

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  # Only allow a trusted parameter "white list" through.
+  def ticket_params
+    params.fetch(:ticket, {}).permit(:contact_id, :number_of_tickets)
```
justintmckibben, line 64: hmm I'm not familiar with fetch

app/controllers/tickets_controller.rb
justintmckibben: The reason why I bring this up is because if we don't expect people to type the dash in themselves, then we shouldn't store the dash on the DB. If we do expect people to type the dash in themselves, then we can go ahead and add it here like it is

nanxiy: i think we shouldn't make them enter the dash, similar to how phone number fields plug in the dashes for you

justintmckibben: FYI @ath-chen is it possible to automatically show the dash in our input? and if they type the dash on the frontend, just strip it before you send it back?

ath-chen: @justintmckibben yeah i can do that, are codes always going to be in a XYZ-123 format (none will be longer / shorter)? To clarify, when it goes to the backend, it'll just look like XYZ123 right?

justintmckibben: I personally think if we do 5 digits, it should be A1C-J1 and if we do 6, then NWK-3JF. What do you think @nanxiy @jacoblb64 

Three digits at a time is easy to remember when typing things in or reading it off to a friend

justintmckibben: @ath-chen FYI this would also mean you'd need to manually input the dash when generating the tickets

jacoblb64: @justintmckibben, I agree with the way you've broken things up, happy with either 5 or 6 digits. I'm pretty sure we're saving with the dash for gift cards, would we want to do the same or different here?

justintmckibben: Let's not save the dashes here — and migrate the gift cards in the future to strip the dashes and hashes. Poor design decision by 3 months ago me 🙃

db/migrate/20200817030307_make_contact_nullable_on_tickets.rb
jacoblb64: Ahaha yeah this was a mistake on our part, but making it better!

app/controllers/tickets_controller.rb
jacoblb64: sure, makes sense to be consistent, I'll swap it

app/controllers/tickets_controller.rb
jacoblb64: Apparently **The two dot form creates an inclusive range, and the three-dot form creates a range that excludes the specified high value** I feel like starting at 1 and using the inclusive range makes a little more sense here? But happy to change it.

http://rubylearning.com/satishtalim/ruby_ranges.html#:~:text=The%20two%20dot%20form%20creates,references%20to%20two%20Fixnum%20objects.

app/controllers/tickets_controller.rb
jacoblb64: OK! Going with 5 digits, no dash in the DB everyone!

app/controllers/tickets_controller.rb
jacoblb64: You're right! Though it wasn't on new, it was on `save` adding the bang now

app/controllers/tickets_controller.rb
jacoblb64: The scaffold template gave this to me? so hesitant to change it honestly

app/controllers/tickets_controller.rb
jacoblb64: It's just a filtering command, though I realized it's actually not needed here, so I'll remove it

app/controllers/tickets_controller.rb
justintmckibben: LOL wow this is so bad Ruby why

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+        end
+      end
+    else
+      savedCorrectly = false
+    end
+
+    if savedCorrectly
+      json_response(createdTickets, :created)
+    else
+      json_response({ message: 'Error: malformed request, expecting participating_seller_id and number_of_tickets' }, :unprocessable_entity)
```
justintmckibben, line 45: Instead you can do 

`raise InvalidParameterError, 'Error: malformed request, expecting participating_seller_id and number_of_tickets'`

and it will be caught by our exception_handller

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+  end
+
+  # PATCH/PUT /tickets/1
+  def update
+    if @ticket.update!(ticket_params)
+      json_response(@ticket)
+    else
+      json_response(@ticket.errors, :unprocessable_entity)
```
justintmckibben, line 54: You don't need this line. It'll be caught by our exception_handler since you have the bang on the update

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+          attributes[:ticket_id] = Ticket.generate_ticket_id
+          attributes[:participating_seller] = @participating_seller
+
+          @ticket = Ticket.new(attributes)
+          savedCorrectly &&= @ticket.save!
+          createdTickets << @ticket
+        end
+      end
+    else
+      savedCorrectly = false
```
justintmckibben, line 39: I still don't think you need this. doing save! should raise an exception that will be caught by our exception_handler

spec/requests/tickets_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+        expect(response.content_type).to match(a_string_including('application/json'))
+      end
+    end
+  end
+
+  describe 'PATCH /update' do
+    context 'with valid parameters' do
+      it 'updates the requested ticket with a valid contact' do
+        ticket = Ticket.create! base_attributes
+        patch ticket_url(ticket),
```
justintmckibben, line 151: We've never used patch before — does put also work?

app/models/ticket.rb
```diff
@@ -27,10 +27,14 @@
...
+  belongs_to :contact, optional: true
   belongs_to :participating_seller
   belongs_to :sponsor_seller, optional: true
 
   validates_presence_of :ticket_id, :participating_seller_id
   validates_uniqueness_of :ticket_id
+
+  def self.generate_ticket_id
+    SecureRandom.alphanumeric(5).upcase
+  end
```
justintmckibben, line 39: You can actually automatiallly generate this in the init if you want

spec/requests/tickets_spec.rb
jacoblb64: They're the same in the rails world apparently, but I'll switch these to put to be more consistent

app/controllers/tickets_controller.rb
jacoblb64: Makes sense to me, switched it all and it works a lot better and is a TON cleaner this way thanks 👍

app/controllers/tickets_controller.rb
jacoblb64: I understand our exception handler better now, removed!

spec/requests/tickets_spec.rb
```diff
@@ -0,0 +1,201 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe '/tickets', type: :request do
```
justintmckibben, line 5: I love these tests! They're super cool — I think I'll follow this format from now on

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    if numTix.blank? || numTix < 1
+      raise InvalidParameterError, 'Error: malformed request, expecting participating_seller_id and number_of_tickets'
+    end
+
+    createdTickets = []
+
+    ActiveRecord::Base.transaction do
+      (1..numTix).each do
+        createdTickets << Ticket.create!({ participating_seller: @participating_seller })
```
justintmckibben, line 32: nit but Ruby is snake_case

app/controllers/tickets_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  # GET /tickets/1
+  def show
+    json_response(@ticket)
+  end
+
+  # POST /tickets
+  def create
+    numTix = params[:number_of_tickets]
```
justintmckibben, line 22: same here — I wonder why rubocop didn't catch this

app/controllers/tickets_controller.rb
jacoblb64: right right right roiiight

spec/requests/tickets_spec.rb
jacoblb64: Ahaha I mostly just followed what the rails scaffolding gave me! The designers know what they're doing 😎

PR #222, opened 2020-08-27
Add a printed boolean to tickets

justintmckibben:
This way we can differentiate new tickets that
we've created that still need to be printed. This
is also useful for when we create "test" tickets
in the prod DB so that we know that we can safely
delete these tickets.

### DIFF
--- a/db/migrate/20200827044947_add_printed_to_tickets.rb
+++ b/db/migrate/20200827044947_add_printed_to_tickets.rb
@@ -1,0 +1,12 @@
+class AddPrintedToTickets < ActiveRecord::Migration[6.0]
+  def change
+    add_column :tickets, :printed, :boolean, null: false, default: false
+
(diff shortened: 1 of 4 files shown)


db/migrate/20200827044947_add_printed_to_tickets.rb
```diff
@@ -0,0 +1,12 @@
+class AddPrintedToTickets < ActiveRecord::Migration[6.0]
+  def change
+    add_column :tickets, :printed, :boolean, null: false, default: false
+
+    # NB(justintmckibben): At the time of running this migration, all of the
+    #                      existing tickets that were created were already
+    #                      printed
+    Ticket.all.each do |ticket|
+      ticket.update(printed: true)
```
jacoblb64, line 9: we should make sure this is really true before merging

PR #223, opened 2020-08-27
Add support for subheader in rewards page

justintmckibben:
https://docs.google.com/document/d/1Cy9qmh-PwKc3P6Q9KPhmsLQ7j5VBzDSDmKixu1HcJi4/edit?disco=AAAAKKBa6to
We didn't notice during the RFC that we could have
two sizes of text inside of the cards for sponor
sellers.


justintmckibben:
@nanxiy added the ability for two different font sizes for sponsor sellers! We'll need to make a follow up PR on the frontend to support each size cc @oliviasztanga 

FYI @Jeremy-Xue we'll need to handle this with our existing data in prod

PR #229, opened 2020-09-01
Sort contact tickets by date

justintmckibben:


app/controllers/contact_tickets_controller.rb
```diff
@@ -7,9 +7,12 @@ class ContactTicketsController < ApplicationController
   def index
     # Don't need the access token for seeing tickets but don't return ticket_id
     # otherwise people can "steal" tickets from other people
-    json_response(Ticket.where(contact: @contact).map do |ticket|
-      ticket.as_json.except(:ticket_id)
-    end)
+    json_response(
+      Ticket.where(contact: @contact)
+        .order(:associated_with_contact_at).map do |ticket|
```
jacoblb64, line 12: Should we specify an order direction? Even if we want the default, it could make it clearer / more readable

spec/requests/contact_tickets_request_spec.rb
```diff
@@ -39,11 +42,12 @@
 
     it 'returns all the tickets this contact has redeemed' do
```
jacoblb64, line 43: maybe update the test name to say "in the correct order" to help debugging in case we ever break this?

PR #231, opened 2020-09-02
Fix undefined exception if email is not provided in param

Yong-L:
Was receiving a bunch of email notification that we couldn't update the contact.


justintmckibben (approved):

jacoblb64:
Yo, mad props for catching and fixing this so quickly! I should have been more diligent myself.

PR #233, opened 2020-09-06
Only return last X tickets

justintmckibben:


justintmckibben:
We need something more robust other than "last". We need true pagination

PR #234, opened 2020-09-07
Pagination and Query Params for Participating Sellers Tickets

jacoblb64:
- Add Pagy for controlling pagination
- Number of items controlled in request (default 50, max 1000)
- Query on either `printed` or `associated_with_contact_at`, to filter items as desired
- Unit tests for the query params, and some basic pagination tests
- Adding pagy metadata to the headers

Followup:
- Once the front-end implements the change (either passthrough of url params or a UI), we can reduce the pagy metadata to what's actually needed

### DIFF
--- a/app/controllers/participating_seller_tickets_controller.rb
+++ b/app/controllers/participating_seller_tickets_controller.rb
@@ -7,9 +7,6 @@ class ParticipatingSellerTicketsController < ApplicationController
 
   # GET /participating_sellers/:participating_seller_id/tickets/:tickets_secret
   def show
-    # NB(justintmckibben): Do we want to hide the tickets that have already
-    #                      been redeemed aka the ones with associated contacts?
-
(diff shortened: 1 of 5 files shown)


app/controllers/participating_seller_tickets_controller.rb
```diff
@@ -1,13 +1,25 @@
 # frozen_string_literal: true
 
+include Pagy::Backend
```
justintmckibben, line 3: I don't think that you need this since it's in the gemfile

justintmckibben (approved):
Looks good! I'm confused by what pagy does though / they aren't covered in tests

app/controllers/participating_seller_tickets_controller.rb
jacoblb64: This is how pagy recommends using functions--as I think they're not scoped but rather recommended to be included

PR #235, opened 2020-09-07
Fix pagy_headers_merge

justintmckibben:


jacoblb64 (changes requested):
If you want to remove the headers we'll have to remove the unit tests too--but this is working fine locally for me right now, did you `bundle install` properly after pulling? I have a feeling if this line if erroring for you, the other pagy methods will too.

jacoblb64:
@justintmckibben, can we close this?

PR #237, opened 2020-09-12
Add Lyft sponsored flag to participating sellers

rjvani:


jacoblb64 (approved):
I'm still a little iffy on the "lyft sponsored" language--but from a code standpoint this is nice and easy done :) We can debate what to call it on the front-end later.

justintmckibben:
You can call it LyftRewardsSeller instead

justintmckibben:
It’s fine as is imo though bc it’s not surfaced anywhere

justintmckibben (approved):

PR #238, opened 2020-09-13
Add giveaway stats to contacts endpoint

justintmckibben:


jacoblb64 (approved):
Looks good to me! I'm a little concerned about performance calculating this on the fly, but I think if we start to see issues later we can refactor.

justintmckibben:
Good callout @jacoblb64. There is an index on contact_id for Ticket so that query should be fast. But you're right it might be worth caching or creating a Redis datastore for things like this. The one we should really revisit is calculating totals raised per merchant and such

PR #239, opened 2020-09-14
Downcasing emails when getting too

jacoblb64:
Current state
- We are already previously downcasing on creating and updating, stopping uppercase pollution in the db

Changes
- Now always downcasing any email param if present to the contacts endpoint
- Added unit tests to check GET and PUT case neutrality

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -69,6 +69,7 @@ class ContactsController < ApplicationController
 
   def set_contact_by_email
     params.require(:email)
+    params[:email] = params[:email].downcase
     @contact = Contact.find_by!(email: params[:email])
   end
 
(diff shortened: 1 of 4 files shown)


justintmckibben (approved):

app/controllers/contacts_controller.rb
```diff
@@ -69,6 +69,7 @@ def update_params
 
   def set_contact_by_email
     params.require(:email)
+    params[:email] = params[:email].downcase
```
justintmckibben, line 72: Could you actually create a separate method called downcase_email that runs before every every method that uses email?

jacoblb64: Sure

jacoblb64: Updated and added unit tests 😉

justintmckibben:
We should be sure to merge accounts with uppercase in them in prod

justintmckibben (approved):

jacoblb64:
> We should be sure to merge accounts with uppercase in them in prod

There's 4 users this is still the case for, one of which is you. Let's strategize the best way to do that :)

PR #240, opened 2020-09-15
Add methods to contact to return Lyft Rewards eligibility

rjvani:
Return `is_eligible_for_lyft_reward` and `has_redeemed_lyft_reward`

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -3,8 +3,6 @@
 class ContactsController < ApplicationController
   before_action :set_contact_by_email, only: %i[index]
   before_action :set_contact_by_id, only: %i[show update]
-  before_action :set_tickets, only: %i[index]
-  before_action :set_lyft_sponsored_participating_sellers, only: %i[index]
 
   # POST /contacts
   def create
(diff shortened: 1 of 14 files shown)


app/controllers/contacts_controller.rb
```diff
@@ -35,15 +35,17 @@ def get_contact_json
     # our DB to become an email -> instagram/name lookup for people to abuse
     ret = @contact.as_json.slice('id')
     ret[:instagram] = @contact.instagram.present?
+    ret[:is_eligible_for_lyft_reward] = @contact.is_eligible_for_lyft_reward
```
danthaman44, line 38: any thoughts on consolidating these two fields? Like just have `is_eligible_for_lyft_reward` be false if `has_redeemed_lyft_reward` is true

danthaman44 (approved):

app/controllers/contacts_controller.rb
rjvani: Yeah, we could definitely do that. Does the UI need `has_redeemed_lyft_reward` at all (do we want to show some badge or some status component)? If not, I can merge these.

justintmckibben: I agree that this might be better handled on the backend. This was outlined as an alt approach here: https://docs.google.com/document/d/1t4J91S21iFQkPrDa9IkUIWcYxoASjZS7B-d3BEwFgfc/edit#heading=h.lk2v4pj3i8vi

justintmckibben (approved):
Nice — great tests!

app/controllers/contacts_controller.rb
rjvani: It currently checks that value first here: https://github.com/sendchinatownlove/ruby/pull/240/files#diff-c4b0c8c9c2b6974a9038fdb3f22c11c4R27. I'm going to hide the flag from the response for now - we can expose it later if needed.

PR #241, opened 2020-09-15
Add endpoint to send e-mails for lyft rewards

rjvani:


justintmckibben (approved):
Wow Rohan you are on fire

PR #243, opened 2020-09-16
Add active boolean to sponsor_sellers

Jeremy-Xue:
Filter out inactive sponsorsellers (yifang)

### DIFF
--- a/db/migrate/20200916000828_add_active_to_sponsor_seller.rb
+++ b/db/migrate/20200916000828_add_active_to_sponsor_seller.rb
@@ -1,5 +1,6 @@
 class AddActiveToSponsorSeller < ActiveRecord::Migration[6.0]
   def change
-    add_column :sponsor_sellers, :active, :boolean
+    add_column :sponsor_sellers, :active, :boolean, default: true
+    change_column_null :sponsor_sellers, :active, false, true
   end
 end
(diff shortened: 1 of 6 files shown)


justintmckibben (approved):

db/migrate/20200916000828_add_active_to_sponsor_seller.rb
```diff
@@ -0,0 +1,5 @@
+class AddActiveToSponsorSeller < ActiveRecord::Migration[6.0]
+  def change
+    add_column :sponsor_sellers, :active, :boolean
```
justintmckibben, line 3: can you add default: true

justintmckibben: and nullable: false

justintmckibben: You can also do the migration here SponsorSellers.all.each { |s| s.update(active: true) }

Jeremy-Xue: just realized this after making the pr, adding now!

Jeremy-Xue: I think change_column_null :sponsor_sellers, :active, false, true also works, by setting nullable to false, and changing all null columns to true

PR #244, opened 2020-09-16
Apply transaction fees

Jeremy-Xue:
We will be charging a 2.75% + 30 cent square transaction fee for every donation, gift card, and pool donation.

- Fees are associated with a campaign, and if no campaign is present, the FeeHandler will automatically apply the square fee.
- Campaigns now has an attribute has_square_fee which, if true, will add the square fee (will be created with id 1) to all new campaigns. 
- flat_cost will be represented as a decimal in the db (and multiplied by 100 when actually being deducted)
- multiplier will be represented as the percent decimal (0.01 is a 1% fee, 0.0275 is a 2.75% fee)

Will be merged in once the frontend is updated to add the square fee to future transactions


app/controllers/campaigns_controller.rb
```diff
@@ -22,6 +22,11 @@ def show
   # POST /campaigns
   def create
     @campaign = Campaign.create!(create_params)
+    if params[:has_square_fee]
+      @fee = Fee.find(1)
+      @campaign.fees << @fee
+      @fee.campaigns << @campaign
```
Jeremy-Xue, line 28: Actually wasn't able to find if this automatically works both ways for has_and_belongs_to_many relationships, I'll write a test to confirm this

db/migrate/20200912021520_create_join_table_campaigns_fees.rb
```diff
@@ -0,0 +1,10 @@
+class CreateJoinTableCampaignsFees < ActiveRecord::Migration[6.0]
+  def change
+    create_join_table :campaigns, :fees do |t|
+      # t.index [:campaign_id, :fee_id]
+      # t.index [:fee_id, :campaign_id]
+    end
+
+    add_column :campaigns, :has_square_fee, :boolean, default: true
```
danthaman44, line 8: Instead of making this a boolean, what if we just made it an optional ID reference to the Fee? If theres no Fee ID, that means there is no fee.

app/controllers/campaigns_controller.rb
```diff
@@ -22,6 +22,11 @@ def show
   # POST /campaigns
   def create
     @campaign = Campaign.create!(create_params)
+    if params[:has_square_fee]
```
danthaman44, line 25: Using the Fee id approach I described, this would become
```
if params[:square_fee_id]
  @fee = Fee.where(:id => :square_fee_id)
```

app/controllers/campaigns_controller.rb
```diff
@@ -22,6 +22,11 @@ def show
...
+      @fee = Fee.find(1)
```
Kenworthe, line 26: small suggestion: move the `1` ("magic number"!) into a const like "SQUARE_FEE_ID" so that it is descriptive. (But might be okay to ignore for now since I guess we only have one Fee on prod lol)

app/services/webhook_manager/donation_creator.rb
```diff
@@ -19,9 +19,13 @@ def call
                                                   seller_id: seller_id,
                                                   payment_intent: payment_intent
                                                 })
+        amount_after_fees = WebhookManager::FeeHandler.call({
+          payment_intent: payment_intent,
+          amount: amount
+        })
```
Kenworthe, line 25: @Jeremy-Xue just wanted to double check this - since .call is an instance method, you might have to initialize with `.new(payment_intent:, amount:)` first?

Kenworthe (changes requested):

Kenworthe:
Looks good to me 👍  just left a few small comments & we should prob delete the commented out lines!

app/controllers/campaigns_controller.rb
jacoblb64: I like this too--easier to scale if/when we have different fee structures too :)

app/controllers/fees_controller.rb
```diff
@@ -11,29 +11,27 @@ def index
...
     @fee.update(update_params)
+    @fee.save
     json_response(@fee)
   end
 
   private
 
   def create_params
-    params.require(:seller_id)
+    #params.require(:campaign_id)
```
jacoblb64, line 27: accidental comments?

app/helpers/sellers_helper.rb
```diff
@@ -4,25 +4,27 @@ module SellersHelper
...
     # for the customer. eg) We might generate a $10 gift card for GAM, but if
     # there is a 10% fee associated with creating that gift card, then we should
     # charge the customer $11.
-    if seller.cost_per_meal.present?
-      json['cost_per_meal'] = seller.cost_per_meal + fees.inject(0) do |_total, fee|
-        (seller.cost_per_meal * fee.multiplier).ceil
-      end
-    end
+
+    # TODO: change this to use fees through campaigns instead
```
jacoblb64, line 22: Are you planning on uncommenting this before checking in?

spec/requests/fees_spec.rb
```diff
@@ -2,31 +2,187 @@
 
 require 'rails_helper'
 
-RSpec.describe 'Fees API', type: :request do
+RSpec.describe FeesController, type: :controller do
```
jacoblb64, line 5: should we move this to `spec/controllers` then?

app/services/webhook_manager/gift_card_creator.rb
```diff
@@ -29,7 +29,11 @@ def call
           recipient: payment_intent.recipient,
           single_use: single_use
         )
-        GiftCardAmount.create!(value: amount, gift_card_detail: gift_card_detail)
+        amount_after_fees = WebhookManager::FeeHandler.call({
```
jacoblb64, line 32: This feels like a weird way to call this utility--instantiating the class then using it's member variables for processing, can we not just define a method that accepts these as arguments?

jacoblb64 (commented):
Looks good so far, but looks like there still might need some work!

app/controllers/campaigns_controller.rb
Jeremy-Xue: this makes sense, thank you! adding to the next commit

Kenworthe (changes requested):

app/controllers/fees_controller.rb
```diff
@@ -8,14 +8,18 @@ def index
     json_response(Fee.all)
   end
 
+    # GET /fees/:name
+    def show
+      json_response(Fee.find(params[:name]))
```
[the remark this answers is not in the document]
Jeremy-Xue: Sounds good, I'll make the changes tonight!

app/controllers/campaigns_controller.rb
```diff
@@ -5,12 +5,7 @@ class CampaignsController < ApplicationController
 
   # GET /campaigns
   def index
-    @campaigns = if params[:active].present?
```
Jeremy-Xue, line 8: these changes were from alice's PR but I'm not sure why they showed up in my files changed? :/

PR #245, opened 2020-09-17
Return 200 success if lyft reward is successfully claimed

danthaman44:
The server needs to return a Success code to the front-end, or the front-end will get an error and show the wrong modal

### DIFF
--- a/app/controllers/contact_lyft_rewards_controller.rb
+++ b/app/controllers/contact_lyft_rewards_controller.rb
@@ -41,6 +41,7 @@ class ContactLyftRewardsController < ApplicationController
         state: 'delivered',
         token: token
       )
+      json_response({})
     end
   end
 


app/controllers/contact_lyft_rewards_controller.rb
```diff
@@ -41,6 +41,7 @@ def create
         state: 'delivered',
         token: token
       )
+      head :ok
```
justintmckibben, line 44: should this be just json_response

justintmckibben (approved):

app/controllers/contact_lyft_rewards_controller.rb
danthaman44: hmm I tried this and it seems `json_response()` requires an object. I could do `json_response(nil)`, but that would essentially be the same thing as `head :ok` except we return a `null` in the body instead of an empty body. My thought is that we should leave the body empty since the front-end doesnt require any data, just the response code.

danthaman44: Actually we can do `json_response({})`, that gives some flexibility if we want to update the response payload later.

PR #248, opened 2020-09-18
Add launch date for lyft rewards

rjvani:


app/models/contact.rb
```diff
@@ -44,6 +44,7 @@ def has_redeemed_lyft_reward
   private
 
   def has_redeemed_lyft_sponsored_ticket
+    lyft_rewards_launch_date = Date.new(2020, 9, 18)
```
danthaman44, line 47: not sure how Date works but maybe we can set 12:100AM as the time to be explicit?

danthaman44 (approved):

app/models/contact.rb
rjvani: I checked in the console, this should be good.

PR #249, opened 2020-09-20
Add check for currently claimed rewards

rjvani:
This PR doesn't allow a User to claim another reward if they currently have a delivered reward which has not expired yet. The unit tests weren't catching this use case properly since we were using `Date` and not a `DateTime`.


justintmckibben (approved):
Ooo good catch

PR #252, opened 2020-09-27
Email edit for final blast

Jeremy-Xue:
Copy: https://docs.google.com/document/d/1gITFmAV24YF9LfqGxMvakt20DfpQt1KqqhiHMGiW8Uk/edit

- Removed bcc to receipts to avoid getting banned by google groups again, will need to update again after sending the blast so we resume sending receipt copies to receipts@

- Will test a few times with different ticket amounts before sending the real blast

- Will be on a flight tomorrow that lands at noon, so should be able to test sending it out around 2

- Test email (0 test tickets in the seedfile)
Screen Shot 2020-09-26 at 9 12 25 PM

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -80,3 +80,6 @@ gem "annotate", "~> 3.1"
 gem 'safe_attributes'
 
 gem 'ulid'
+
+# pagination gem
+gem 'pagy', '~> 3.8'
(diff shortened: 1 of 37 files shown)


nanxiy:
Can we add an extra line break before the “You’re not entered...” line?
Break it up a bit since it feels a bit weird that we have 6 sentences all
spaced out. Also when you test w diff amounts can you make sure that the
Grand prize line shows up :)

PR #257, opened 2020-09-27
Wilsonj806/containerization instructions update

wilsonj806:


justintmckibben (approved):
Thanks for doing this!!

README.md
```diff
@@ -109,24 +108,24 @@ https://docs.google.com/document/d/1UPNCwjWS_T7XT5AXsewphu6NvNdV7TQLSJub-RBRAG0/
...
-cd api && docker-compose up
-# navigate to localhost:3000
+cd <scl-be directory> && docker-compose up
 ```
 accessing rail CLI after `docker-compose` is running.
 
 ```
-docker exec -it api_web_1 bash
-# now inside docker container example of commands that can be run
+# Replace <scl-be directory name> with the name of the direcotry that you cloned the repo into
```
jacoblb64, line 120: maybe add a comment that a user can confirm the image name with `docker-compose ps`?

config/database.yml
```diff
@@ -29,6 +29,8 @@ development:
   # To create additional roles in postgres see `$ createuser --help`.
   # When left blank, postgres will use the default role. This is
   # the same name as the operating system user that initialized the database.
+  # NOTE: You will need to uncomment  and use the `username` value that corresponds with
+  # ...the POSTGRES_USER value in docker-compose.yml
```
jacoblb64, line 33: just to confirm -- should these be uncommented for local or docker running of the database?

jacoblb64 (approved):
Thank you so much for doing this! Left some small possibly clarifying comments if you want, but they're completely optional.

config/database.yml
wilsonj806: Oh yeah, I forgot to specify that. Updated with that in mind and your above comment about `docker-compose ps`

jacoblb64: Awesome, thanks for taking care of this again and and clarifying the instructions!

PR #259, opened 2020-10-04
Added nonprofit id to campaign model

danthaman44:
Add nonprofit id reference to the campaign.
Test plan:
Screen Shot 2020-10-03 at 10 53 19 PM

```
danthaman-mac:ruby danthaman$ heroku local:run bundle exec rspec spec/models/campaign_spec.rb
.....
Finished in 0.46409 seconds (files took 1.7 seconds to load)
5 examples, 0 failures

danthaman-mac:ruby danthaman$ heroku local:run bundle exec rspec spec/requests/campaigns_request_spec.rb
......................
Finished in 1.26 seconds (files took 1.9 seconds to load)
22 examples, 0 failures
```

### DIFF
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -45,12 +45,14 @@ class CampaignsController < ApplicationController
       :end_date,
       :price_per_meal,
       :target_amount,
+      :nonprofit_id,
       gallery_image_urls: []
     )
 
     set_location
     set_seller
     set_distributor
+
     ret[:location_id] = @location.id
     ret[:seller_id] = @seller.id
     ret[:distributor_id] = @distributor.id
@@ -64,6 +66,7 @@ class CampaignsController < ApplicationController
       :description,
       :valid,
       :price_per_meal,
+      :nonprofit_id,
       gallery_image_urls: []
     )
   end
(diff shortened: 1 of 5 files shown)


db/migrate/20201004014328_add_nonprofit_id_to_campaigns.rb
```diff
@@ -0,0 +1,5 @@
+class AddNonprofitIdToCampaigns < ActiveRecord::Migration[6.0]
+  def change
+    add_column :campaigns, :nonprofit_id, :integer
```
justintmckibben, line 3: This should also have a foreign key and index

justintmckibben: add_reference : campaigns, :nonprofit, index: true

app/controllers/campaigns_controller.rb
```diff
@@ -45,12 +45,17 @@ def create_params
...
       gallery_image_urls: []
     )
 
     set_location
     set_seller
     set_distributor
+    if params[:nonprofit_id].present?
+      set_nonprofit
+      ret[:nonprofit_id] = @nonprofit.id
+    end
```
justintmckibben, line 58: We shouldn't need this since it'll be on the campaign itself

db/migrate/20201004014328_add_nonprofit_id_to_campaigns.rb
danthaman44: Did not know about reference column! Thanks for sharing

PR #260, opened 2020-10-07
Add additional seeds for OpenHours, DeliveryOptions, and DeliveryTypes

wilsonj806:


db/seeds.rb
```diff
@@ -62,6 +62,121 @@
   Seller.find_or_initialize_by(seller_id: attributes[:seller_id]).update!(attributes)
 end
 
+[
+    {
+        seller_id: 1,
+        open_time: '2000-01-01T07:00:00.000Z',
```
justintmckibben, line 68: It's strange to me that open time and closing times have dates associated — that probably shouldn't be the case on the model

justintmckibben (approved):

db/seeds.rb
wilsonj806: Yeah I found that strange as well, especially since I had to do some funky formatting to convert from the date format to a usable time.

justintmckibben: Mind adding a ticket into Trello to update the model to only be time and not datetime?

wilsonj806: Yup, which column should it be under?

wilsonj806: Also fixed the seed file so that it populates the database with just the time, and not the full time. I still get the full date time in the response though.

justintmckibben: Yeah that's weird. I'd run a migration on that column to only store the time

wilsonj806: Don't think it's that as my schema.rb file seems fine and running an SQL query on it returns the expected time. Looking into it, it looks like it's a quirk with Ruby's implementation of the Time class where it's storing the seconds since the Epoch.

PR #261, opened 2020-10-11
Return GAM in seller model

chrisng93:
Return GAM in seller model. Right now, we return the gift card amount for the seller, which includes both gift card donations and GAM. We might want to break this out in the future to have amounts for GAM and non-GAM gift cards (not sure the best name for the latter). This is my first time writing Ruby/RoR so feel free to give me any suggestions on my code

Trello ticket: https://trello.com/c/2Ncvuvix/609-add-gam-in-total-raised-contribution-bar-on-merchants-page


chrisng93:
not 100% what's going on with the tests. looks like they're failing on an environment variable not being there. i ran tests locally and everything passed. anyone have an idea what's going on?

PR #262, opened 2020-10-15
Return GAM in seller model (Chris' change)

jacoblb64:
Return GAM in seller model. Right now, we return the gift card amount for the seller, which includes both gift card donations and GAM. We might want to break this out in the future to have amounts for GAM and non-GAM gift cards (not sure the best name for the latter). This is my first time writing Ruby/RoR so feel free to give me any suggestions on my code

Trello ticket: https://trello.com/c/2Ncvuvix/609-add-gam-in-total-raised-contribution-bar-on-merchants-page

### DIFF
--- a/app/helpers/sellers_helper.rb
+++ b/app/helpers/sellers_helper.rb
@@ -25,6 +25,7 @@ module SellersHelper
     end
 
     json['donation_amount'] = seller.donation_amount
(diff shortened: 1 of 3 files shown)


danthaman44 (approved):
LGTM!

PR #263, opened 2020-10-19
Adding some of the one-off scripts that we've run for visibility

jacoblb64:
- Making a place for us to keep track of these one-off operations as we seem to be doing more of them
- This will help us see what we've done, learn how we did them, and perform future steps too.

### DIFF
--- a/scripts/break-up-1000-giftcard.rb
+++ b/scripts/break-up-1000-giftcard.rb
@@ -1,0 +1,57 @@
+Item.transaction do
+    # a priori
+    big_card_id = 9826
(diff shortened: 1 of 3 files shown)


justintmckibben (approved):
I wonder if migration scripts are the right place to store these?

jacoblb64:
> I wonder if migration scripts are the right place to store these?

I was thinking about that or rake tasks--but these aren't really either of those and I don't want to confuse them. I think these are mostly historical documentation for us as a team. I think there's definitely a better top-level folder name than `scripts`, but I'm not sure what that would be 🤔.

PR #264, opened 2020-10-22
Add gift_card_id and last_updated to gift card return in seller gift card endpoint

chrisng93:
Add gift_card_id and last_updated to gift card return in seller gift card endpoint. These pieces of data are needed for rendering the correct gift card info and updating gift card info in the merchant dashboard v2

### DIFF
--- a/app/controllers/seller_gift_cards_controller.rb
+++ b/app/controllers/seller_gift_cards_controller.rb
@@ -17,7 +17,7 @@ class SellerGiftCardsController < ApplicationController
               :updated_at,
               :expiration,
               :single_use,
-              'la.created_at as last_updated'
+              'la.updated_at as last_updated'
             )
             .joins(:item, :recipient)
             .where(
--- a/app/controllers/seller_gift_cards_controller.rb
+++ b/app/controllers/seller_gift_cards_controller.rb
@@ -7,6 +7,7 @@ class SellerGiftCardsController < ApplicationController
   def show
     query = GiftCardDetail
             .select(
+              :gift_card_id,
               :seller_gift_card_id,
               'la.value as latest_value',
               'og.value as original_value',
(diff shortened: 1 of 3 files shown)


chrisng93:
Tests failing again because of the secrets in my fork not being set up (`Validation failed: Square location can't be blank`). @justintmckibben any idea on how to get the tests to pass on a fork?

app/models/gift_card_detail.rb
```diff
@@ -43,6 +43,13 @@ def amount
                   .value
   end
 
+  def last_updated
```
jacoblb64, line 46: Is this being called anywhere right now?

spec/requests/seller_gift_cards_request_spec.rb
```diff
@@ -44,6 +44,7 @@
 
       let(:expected_gift_card_latest_amount) { 30_00 }
       let(:expected_gift_card_original_amount) { 10_00 }
+      let(:expected_gift_card_last_updated) { Time.current + 1.day }
```
jacoblb64, line 47: nit: shouldn't this time be in the past? Not that it matters for a test 😂

app/controllers/seller_gift_cards_controller.rb
```diff
@@ -15,7 +16,8 @@ def show
               :created_at,
               :updated_at,
               :expiration,
-              :single_use
+              :single_use,
+              'la.created_at as last_updated'
```
jacoblb64, line 20: I know that these are pretty much always the same as we create a new row for each redemption, but do you think there's any value here in using `updated_at` just in case someone does manually update a record in the db?

jacoblb64 (approved):
Looks good man! Left some small non-blocking comments

spec/requests/seller_gift_cards_request_spec.rb
chrisng93: yeah that's fair. it's easier to do it this way though because the created_at/updated_at for the other gift cards are now(). so to make this row a current/past time, you'd have to update the created_at/updated_at for the rest of the other gift cards to be a time before that so that this row is the newest

PR #265, opened 2020-10-23
Add gift_card_id and last_updated to gift card return in seller gift card endpoint

chrisng93:
Add gift_card_id and last_updated to gift card return in seller gift card endpoint. These pieces of data are needed for rendering the correct gift card info and updating gift card info in the merchant dashboard v2

Dupe of https://github.com/sendchinatownlove/ruby/pull/264 but not from a fork


jacoblb64 (approved):
Looks good to me! Thanks for addressing my comments from the dupe too :)

PR #267, opened 2020-11-02
Draft: Payment for Projects

jacoblb64:
Looking for some feedback and some help testing with this!

Work Completed
- Handling both projects and sellers for the /charges endpoint
- Dealing with all the logic for webhooks
- Misc. Migrations annotations that were automatically updated

Testing Done
- Testing simple happy path for a payment with a project ID and no seller ID

Outstanding Work
- Update unit tests
- Test all existing functionality

### DIFF
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -123,7 +123,8 @@ class WebhooksController < ApplicationController
 
       amount = item_json['amount']
       seller_id = item_json['seller_id']
-      if seller_id.eql?(Seller::POOL_DONATION_SELLER_ID)
+      project_id = item_json['project_id']
+      if seller_id.present? && seller_id.eql?(Seller::POOL_DONATION_SELLER_ID)
         PoolDonationValidator.call({ type: item_json['item_type'] })
 
         WebhookManager::PoolDonationCreator.call(
@@ -138,6 +139,24 @@ class WebhooksController < ApplicationController
           {
             payment_intent: payment_intent,
             amount: amount,
+            email: payment_intent.purchaser.email
+          }
+        )
+      elsif project_id.present?
+        WebhookManager::DonationCreator.call(
+          {
+            seller_id: seller_id,
+            project_id: project_id,
+            payment_intent: payment_intent,
+            amount: amount
+          }
+        )
+
+        EmailManager::DonationReceiptSender.call(
+          {
+            payment_intent: payment_intent,
+            amount: amount,
+            merchant: Project.find(project_id).name,
             email: payment_intent.purchaser.email
           }
         )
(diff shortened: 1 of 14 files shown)


app/controllers/webhooks_controller.rb
```diff
@@ -193,14 +212,16 @@ def handle_payment_intent_succeeded(
...
       grouped_items.each_key do |sid|
-        EmailManager::DonationReceiptSender.call(
-          {
-            payment_intent: payment_intent,
-            amount: grouped_items[sid].map { |li| li['amount'].to_f }.sum,
-            merchant: Seller.find_by(seller_id: sid).name,
-            email: payment_intent.purchaser.email
-          }
-        )
+        if sid.present?
```
justintmckibben, line 215: This means that we don't send emails for light up chinatown donatoins?

jacoblb64, line 215: as seller_id can now be null

app/services/webhook_manager/item_creator.rb
```diff
@@ -3,23 +3,35 @@
...
+          campaign_id: payment_intent.campaign_id
+        )
+      else
+        Item.create!(
+          project: project,
+          purchaser: payment_intent.purchaser,
+          item_type: item_type,
+          payment_intent: payment_intent
+        )
+      end
```
justintmckibben, line 34: I wonder if we should raise if neither seller nor project exist

app/controllers/webhooks_controller.rb
```diff
@@ -141,6 +142,24 @@ def handle_payment_intent_succeeded(
...
+        WebhookManager::DonationCreator.call(
+          {
+            seller_id: seller_id,
+            project_id: project_id,
+            payment_intent: payment_intent,
+            amount: amount
+          }
+        )
+
+        EmailManager::DonationReceiptSender.call(
```
jacoblb64, line 155: not setting `is_donation` and instead sending a one-off email from here--could be customized further too

justintmckibben (approved):
Looks good! Nothing blocking

app/services/webhook_manager/item_creator.rb
jacoblb64: Was gonna mention the same thing, for this and some other places, if we should have a default case where neither is there and throw an exception

app/controllers/webhooks_controller.rb
jacoblb64: `seller_id` can be null now, when there is a `project_id`, so we should skip it here (the grouping logic seemed weird, and it should be on it's own anyway). I'm triggering a separate email in the case above for a project

app/services/webhook_manager/item_creator.rb
Jeremy-Xue: tiny bit of overlap with what I was working on here, but this is basically what I had!

PR #268, opened 2020-11-02
Wilson/add open hours delivery option seed data

wilsonj806:


db/seeds.rb
```diff
@@ -228,3 +228,122 @@
...
+      icon_url: './assets/Caviar@2x.png'
+    }
+  },
+  {
+    id: 3,
+    url: 'http://doordash.com/menu/',
+    seller_id: 1,
+    delivery_type: {
+      id: 3,
+      name: 'Door Dash',
```
justintmckibben, line 295: nit but I'm pretty sure DoorDash is upper camel-cased

justintmckibben (approved):

db/seeds.rb
wilsonj806: you're right, fixed!

PR #269, opened 2020-11-04
Remove active parameter in Get All Campaigns

aliceluo2048:
I don't think we are passing in an `active` query in the get all `Campaigns` controller (from what I can see in the front end).  When would this be used? I suggest we remove this and sort the GAM campaigns on the front end.

### DIFF
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -5,12 +5,7 @@ class CampaignsController < ApplicationController
 
   # GET /campaigns
   def index
-    @campaigns = if params[:active].present?
-                   valid_campaigns.order(:end_date).active(params[:active])
-                 else
-                   valid_campaigns.order(:end_date).all
-    end
-
+    @campaigns = valid_campaigns.order(:end_date).all
     json_response(campaigns_json)
   end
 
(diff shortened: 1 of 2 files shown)


justintmckibben (approved):
I agree that you can remove the active flag, but I think that we should be sorting on the backend. The reason is because what if in the future we don't return all of the campaigns, and instead we paginate. If we sort on the frontend, we'll have to refactor

aliceluo2048:
> I agree that you can remove the active flag, but I think that we should be sorting on the backend. The reason is because what if in the future we don't return all of the campaigns, and instead we paginate. If we sort on the frontend, we'll have to refactor

The front end change is just adding `.reverse()` as the front end is already filtering whether if it's active or not. If we need to add pagination, we can just delete this reverse  and add the more complex sorting in the back end.

justintmckibben:
> > I agree that you can remove the active flag, but I think that we should be sorting on the backend. The reason is because what if in the future we don't return all of the campaigns, and instead we paginate. If we sort on the frontend, we'll have to refactor
> 
> The front end change is just adding `.reverse()` as the front end is already filtering whether if it's active or not. If we need to add pagination, we can just delete this reverse and add the more complex sorting in the back end.

The frontend should probably use the active filter, and sort. In general frontends should be as dumb as possible and we should frontload any sorting, filtering or calculations on the backend where possible.

PR #271, opened 2020-11-06
Payments for Projects (take 2)

jacoblb64:
This is a second attempt at #267, after the accidental premature merge to master was reverted in #270.

Work Completed
- Handling both projects and sellers for the /charges endpoint
- Dealing with all the logic for webhooks
- Misc. Migrations annotations that were automatically updated
- Updating existing tests to work with new changes
- Adding new test cases to cover new use cases with Projects

### DIFF
--- a/spec/requests/charges_spec.rb
+++ b/spec/requests/charges_spec.rb
@@ -10,7 +10,7 @@ RSpec.describe 'Charges API', type: :request do
     let(:is_square) { true }
     let(:name) { 'Squarepants, Spongebob' }
     let(:seller_id) { 'shunfa-bakery' }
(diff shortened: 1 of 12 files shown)


spec/requests/charges_spec.rb
```diff
@@ -190,6 +191,56 @@
         end
       end
 
+      context 'with a project_id' do
```
justintmckibben, line 194: What happens if it has both?

justintmckibben (approved):

spec/requests/charges_spec.rb
Jeremy-Xue: Just added the test, works as expected by returning a 422 error.

I think it should be good to go!

PR #274, opened 2020-11-09
edited the distributors table in the schema to account for tokens

ryannamgung:


rjvani (changes requested):
Can you add these changes in a migration file?

PR #280, opened 2020-11-12
Return seller/distributor pairings for campaign

chrisng93:
Return seller/distributor pairings for campaign. The return array is nested within the campaign under the 'seller_distributor_pairs' field. Each seller_distributor_pair consists of id, image_url, and name for the seller and distributor

If campaign has no seller/distributor pairings:
Screen Shot 2020-11-11 at 8 18 41 PM

If campaign has pairings:
Screen Shot 2020-11-11 at 8 18 53 PM

Ref: https://trello.com/c/F0FanzWN/703-megagam-update-the-campaigns-api-to-return-sellers-and-distributors-lists-pairs

### DIFF
--- a/app/models/campaign.rb
+++ b/app/models/campaign.rb
@@ -55,6 +55,46 @@ class Campaign < ApplicationRecord
     ).order(created_at: :desc).first&.created_at
   end
 
+  def seller_distributor_pairs
+    pairs = []
+
+    csd_pairings = CampaignsSellersDistributor
+      .joins(:campaign)
+      .where(campaigns: {
+        id: id
+      })
+
+    seller_ids = csd_pairings.map {|pairing| pairing.seller_id}
+    sellers = Seller.where(:id => seller_ids)
+    seller_id_to_seller = {}
+    sellers.each do |seller|
+      seller_id_to_seller[seller.id] = seller
+    end
+
+    distributor_ids = csd_pairings.map {|pairing| pairing.distributor_id}
+    distributors = Distributor.where(:id => distributor_ids)
+    distributor_id_to_distributor = {}
+    distributors.each do |distributor|
+      distributor_id_to_distributor[distributor.id] = distributor
+    end
+
+    csd_pairings.each do |pairing|
+      seller = seller_id_to_seller[pairing.seller_id]
+      distributor = distributor_id_to_distributor[pairing.distributor_id]
+      pair = {
+        'distributor_id' => distributor.id,
+        'distributor_image_url' => distributor.image_url,
+        'distributor_name' => distributor.name,
+        'seller_id' => seller.id,
+        'seller_image_url' => seller.hero_image_url,
+        'seller_name' => seller.name,
+      }
+      pairs.append(pair)
+    end
+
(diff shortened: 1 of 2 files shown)


app/models/campaign.rb
```diff
@@ -55,6 +55,46 @@ def last_contribution
     ).order(created_at: :desc).first&.created_at
   end
 
+  def seller_distributor_pairs
```
chrisng93, line 58: does anyone know a better way to do this in rails? i'm trying to do something in 1 query, but i couldn't find a way... i'd need something like:

```
    CampaignsSellersDistributor
      .find_by_sql(
        [
          """
          SELECT
            csd.id,
            csd.campaign_id,
            csd.seller_id,
            st.name as seller_name,
            s.hero_image_url as seller_image,
            csd.distributor_id,
            d.name as distributor_name,
            d.image_url as distributor_image
          FROM campaigns_sellers_distributors csd
          JOIN campaigns c ON csd.campaign_id=c.id
          JOIN sellers s ON csd.seller_id=s.id
          JOIN seller_translations st ON st.seller_id=csd.seller_id
          JOIN distributors d ON csd.distributor_id=d.id
          WHERE csd.campaign_id=? AND st.locale=?
          """,
          id,
          ???,
        ]
      )
```

chrisng93: also please comment on my ruby code i don't know any ruby so i'm pretty sure this code is trash

app/models/campaign.rb
```diff
@@ -55,6 +55,46 @@ def last_contribution
...
+    end
+
+    distributor_ids = csd_pairings.map {|pairing| pairing.distributor_id}
+    distributors = Distributor.where(:id => distributor_ids)
+    distributor_id_to_distributor = {}
+    distributors.each do |distributor|
+      distributor_id_to_distributor[distributor.id] = distributor
+    end
+
+    csd_pairings.each do |pairing|
```
justintmckibben, line 81: it's technically more restful to have the caller of this endpoint make separate calls to the sellers and distributors endpoint to find this stuff

app/models/campaign.rb
justintmckibben, line 58: nit: mind writing a test? You can defer til we decide on the API though

justintmckibben (approved):

app/models/campaign.rb
chrisng93: yeah that's true but it adds complexity on the frontend and doesn't add a ton of latency for most of the campaigns. i can see the case for having clients call the api again to get sellers/distributors but in the case of campaigns, anywhere we're getting them we're showing the seller/distributor info anyways. i think having it all in 1 endpoint simplifies the frontend and doesn't unnecessarily fetch seller/distributor data for any other use case. in this case i'd prefer complexity on the backend vs frontend

app/models/campaign.rb
chrisng93: yeah sure happy to write a test. i might merge this first but can we pair on it? i've been trying to write one but i have no idea where any of the variables are coming from

PR #282, opened 2020-11-12
Comment on payment_intents.successful

chrisng93:
did some thinking about the payments wrt mega gam and wanted get thoughts. so right now we set the payment_intent to be successful when we create some sort of item. i think one way we can think about the payment_intent.successful field is that it represents that the payment has been acknowledged by our system, aka we've received it successfully and done what we need to do with it / square doesn't need to it through again. with this thinking, we don't need to make any changes to the existing WebhookManagers. all we need to do is set the payment_intent to be successful for payment_intents related to mega gam campaigns in the webhook controller

### DIFF
--- a/app/models/payment_intent.rb
+++ b/app/models/payment_intent.rb
@@ -9,6 +9,11 @@
 #  lock_version       :integer
 #  metadata           :text
 #  receipt_url        :string
+#  successful represents whether a payment has been received and processed by
+#  our system. For pooled donations and normal GAM campaigns, this means that
+#  we've created a PooledDonation, Donation, or GiftCard using the payment
+#  from Square. For Mega GAM campaigns, this means that we've received some
+#  sort of ack from Square.
 #  successful         :boolean          default(FALSE)
 #  created_at         :datetime         not null
 #  updated_at         :datetime         not null


app/models/payment_intent.rb
```diff
@@ -9,6 +9,11 @@
 #  lock_version       :integer
 #  metadata           :text
 #  receipt_url        :string
+#  successful represents whether a payment has been received and processed by
```
chrisng93, line 12: this file is autogenerated but i have no idea how to make a comment on a column in rails. if anyone has any good way to do this, lmk

justintmckibben (approved):

chrisng93:
@justintmckibben @jacoblb64 @rjvani @Jeremy-Xue @aliceluo2048  wanted to get your thoughts on this

chrisng93:
can someone point me to relevant square docs btw? i figure we should be catching some sort of unprocessed error in the webhook as well. right now we can't tell the difference between "haven't heard back from square yet" and "there was an error processing the payment"

jacoblb64:
@chrisng93 
> can someone point me to relevant square docs btw? i figure we should be catching some sort of unprocessed error in the webhook as well. right now we can't tell the difference between "haven't heard back from square yet" and "there was an error processing the payment"

We mostly use the `Order` type webhooks
https://developer.squareup.com/docs/webhooks-api/subscribe-to-events#payments

app/models/payment_intent.rb
jacoblb64: Looks like you can add a comment to the schema itself, which should get pulled to this autogenerated comment
https://blog.bigbinary.com/2016/06/21/rails-5-supports-adding-comments-migrations.html

chrisng93:
@jacoblb64 thanks! i'm going to hold off on merging this. i don't love the comments thing for rails

PR #283, opened 2020-11-13
Revert "Revert "Add projects endpoint""

justintmckibben:
This reverts commit 14752e56a43c188d0064085954f1a156c03b56f4.

I accidentally force pushed to master, so reverted it, then this is the unrevert. 

Original commit message:
>Add projects endpoint

> Also add amount_raised to model and amount to
>payment_intent model

### DIFF
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -78,7 +78,7 @@ class CampaignsController < ApplicationController
 
     ret = params.permit(
       :distributor_id,
-      :seller_id,
+      :seller_id
     )
 
     set_campaign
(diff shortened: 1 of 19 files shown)


danthaman44 (approved):

jacoblb64 (approved):
Killin' it man, this looks great!

Quick note: I don't think it affects us, but I've had issues with revert-reverts in the past (specifically moving between dev, release-candidate, and master branches). I usually `cherry-pick --no-commit` the original commit instead!

justintmckibben:
> Killin' it man, this looks great!
> 
> Quick note: I don't think it affects us, but I've had issues with revert-reverts in the past (specifically moving between dev, release-candidate, and master branches). I usually `cherry-pick --no-commit` the original commit instead!

Will do this next time!

PR #285, opened 2020-11-13
Add start_date to campaigns

aliceluo2048:
startdate
https://trello.com/c/tcbvA8xU/690-add-start-date-to-campaigns-model
Also removed inactive campaigns in campaigns_request tests which I forgot to clean up last PR


justintmckibben (approved):

justintmckibben:
not for this PR, but I wonder if we should not return a campaign if it's before the current start date 🤔

PR #286, opened 2020-11-13
Add project_id to campaigns for Mega-GAM

Jeremy-Xue:
Add project_id and make seller_id optional for campaigns, such that project_id xor seller_id must exist for a campaign.

This was added to allow us to associate a GAM campaign with a project which will be used by Mega-GAM

### DIFF
--- a/db/migrate/20201113001940_add_project_id_to_campaigns.rb
+++ b/db/migrate/20201113001940_add_project_id_to_campaigns.rb
@@ -1,0 +1,7 @@
+class AddProjectIdToCampaigns < ActiveRecord::Migration[6.0]
+  def change
+    add_column :campaigns, :project_id, :bigint, null: true
+    add_index :campaigns, :project_id
+    add_foreign_key :campaigns, :projects
+  end
+end
--- a/spec/requests/campaigns_request_spec.rb
+++ b/spec/requests/campaigns_request_spec.rb
@@ -5,11 +5,13 @@ require 'rails_helper'
 RSpec.describe 'Campaigns API', type: :request do
   before do
     @seller = create :seller
+    @project = create :project
     @location = create(:location, seller_id: @seller.id)
     @campaign = create(
       :campaign,
       active: true,
       seller_id: @seller.id,
+      project_id: nil,
       location_id: @location.id
     )
   end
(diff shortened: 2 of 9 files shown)


spec/requests/campaigns_request_spec.rb
```diff
@@ -171,6 +181,35 @@
...
+
+            campaign = Campaign.find(response_body['id'])
+            expect(campaign).not_to be_nil
+            expect(campaign.location).to eq @location
+            expect(campaign.project).to eq @project
+            expect(campaign.target_amount).to eq 100_000
+            expect(campaign.amount_raised).to eq 0
+            expect(campaign.price_per_meal).to eq 500
+
+            expect(campaign.active).to eq false
```
justintmckibben, line 209: I think to be_true/be_false is a thing

justintmckibben (approved):

db/migrate/20201113001940_add_project_id_to_campaigns.rb
```diff
@@ -0,0 +1,7 @@
+class AddProjectIdToCampaigns < ActiveRecord::Migration[6.0]
+  def change
+    add_column :campaigns, :project_id, :bigint, null: true
+    add_index :campaigns, :project_id
+    add_foreign_key :campaigns, :projects
```
justintmckibben, line 5: fyi `add_reference :campaigns, :project, null: true, foreign_key: true` does the same thing all in one line

spec/requests/campaigns_request_spec.rb
```diff
@@ -171,6 +181,35 @@
             expect(campaign.valid).to eq true
           end
         end
+
+        context 'with valid parameters and project_id' do
```
justintmckibben, line 185: we don't have a case that tests creating with both a seller and a project

chrisng93 (approved):

spec/requests/campaigns_request_spec.rb
Jeremy-Xue: changed to 'expect( ).to be true' since I think be_true checks for truthiness vs the actual value

spec/requests/campaigns_request_spec.rb
Jeremy-Xue: added!

PR #288, opened 2020-11-15
Add/update methods on campaign model to get amount allocated/raised for mega GAM campaigns

chrisng93:
Add/update methods on campaign model to get amount allocated/raised for mega GAM campaigns

Ref: https://trello.com/c/G6DvJnWK/705-megagam-create-allocatedamount-calculated-field-for-mega-gam and https://trello.com/c/HGoRxNWg/704-megagam-create-totaldonationamount-calculated-field-for-mega-gam

### DIFF
--- a/app/models/campaign.rb
+++ b/app/models/campaign.rb
@@ -112,18 +112,17 @@ class Campaign < ApplicationRecord
   private
 
   def mega_gam?
-    project_id != nil
+    project.present?
   end
 
   # Calculates the amount raised in the payment intent table.
   def payment_intent_amount
-    # TODO CHECK SYNTAX
     PaymentIntent
       .joins(:campaign)
       .where(campaigns: {
         campaign_id: id
       })
-      .map {$.amount}
+      .map { |payment_intent| payment_intent.amount }
       .sum
   end
 
--- a/app/models/campaign.rb
+++ b/app/models/campaign.rb
@@ -40,10 +40,23 @@ class Campaign < ApplicationRecord
   scope :active, ->(active) { where(active: active) }
 
   def amount_raised
-    # NB(justintmckibben): Currently campaigns are designed only for GAM and
-    # therefore only create gift cards. For campaigns where we don't need the
-    # gift cards aka we distribute hot meals, we *could* create donations
-    # instead of gift cards
+    if mega_gam?
+      # For Mega GAM campaigns, we don't generate items right away, so we rely
+      # on the payment_intents table to determine the total amount raised.
+      payment_intent_amount
+    else
+      # NB(justintmckibben): Currently campaigns are designed only for GAM and
+      # therefore only create gift cards. For campaigns where we don't need the
+      # gift cards aka we distribute hot meals, we *could* create donations
+      # instead of gift cards
+      gift_card_amount
+    end
+  end
+
+  def amount_allocated
+    # Amount allocated is always equal to the gift card amount. We allocate
+    # payments immediately for regular GAM. For Mega GAM, we allocate after
+    # the campaign is over.
     gift_card_amount
   end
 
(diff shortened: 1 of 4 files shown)


app/models/campaign.rb
```diff
@@ -104,6 +118,19 @@ def seller_distributor_pairs
...
 
+  def mega_gam?
+    project.present?
+  end
+
+  # Calculates the amount raised in the payment intent table.
+  def payment_intent_amount
+    PaymentIntent
+      .joins(:campaign)
+      .where('campaigns.id=? AND successful=true', id)
```
justintmckibben, line 129:
```.where(id: id)
.where(successfull: true)```

File: app/models/campaign.rb [hunk 2]
```diff
@@ -104,6 +118,19 @@ def seller_distributor_pairs
...
+      .map { |payment_intent| payment_intent.amount }
```
justintmckibben, line 130: nit but you can do:
`.map { &.amount }`

spec/models/campaign_spec.rb
```diff
@@ -115,18 +116,28 @@
...
+  
+  context 'with amount raised for mega gam campaigns' do
+    let!(:campaign) { create(:campaign, :with_sellers_distributors, :with_project, seller: nil) }
+
+    it 'returns payment intent amounts' do
+      # Expect these 2 to be counted. :with_line_items adds line items of value 600.
+      payment_intent_1 = create(:payment_intent, :with_line_items, campaign: campaign, successful: true)
+      payment_intent_2 = create(:payment_intent, :with_line_items, campaign: campaign, successful: true)
+      # Do not expect this one to be counted since it's not successful.
+      payment_intent_3 = create(:payment_intent, :with_line_items, campaign: campaign, successful: false)
```
justintmckibben, line 133: mind adding these into a subject?
```
  context 'with amount raised for mega gam campaigns' do
    let!(:campaign) { create(:campaign, :with_sellers_distributors, :with_project, seller: nil) }
    subject do 
      # Expect these 2 to be counted. :with_line_items adds line items of value 600.
      payment_intent_1 = create(:payment_intent, :with_line_items, campaign: campaign, successful: true)
      payment_intent_2 = create(:payment_intent, :with_line_items, campaign: campaign, successful: true)
      # Do not expect this one to be counted since it's not successful.
      payment_intent_3 = create(:payment_intent, :with_line_items, campaign: campaign, successful: false)
    end

    it 'returns payment intent amounts' do
      subject 
      expect(campaign.amount_raised).to eq 1200
    end
  end
```

justintmckibben: or inside of a 
```
before do
  ... 
end
```

app/models/campaign.rb
```diff
@@ -104,6 +118,19 @@ def seller_distributor_pairs
 
   private
 
+  def mega_gam?
```
justintmckibben, line 121: No need to do this now, but this attribute probably will be public in the future

spec/models/campaign_spec.rb
justintmckibben: or inside of a method in this context
```
def create_mega_gam_payment_intents
   ...
end
it 'returns payment intent amounts' do
  create_mega_gam_payment_intents
  ...
```

justintmckibben (approved):

app/models/campaign.rb
chrisng93: yeah i'm going to address this in https://github.com/sendchinatownlove/ruby/pull/290/files

app/models/campaign.rb
chrisng93: that syntax isn’t working for me:
```SyntaxError:
  /Users/chrisng/dev/github/ruby/app/models/campaign.rb:133: syntax error, unexpected &.
        .map { &.amount }```

app/models/campaign.rb
chrisng93: done

spec/models/campaign_spec.rb
chrisng93: done

PR #290, opened 2020-11-16
Update Square webhook to *not* create any item for mega GAM campaigns

chrisng93:
Update Square webhook to *not* create any item for mega GAM campaigns. We should just be setting these payment intents to be successful

Ref: https://trello.com/c/KBhOqbJa/695-megagam-update-the-square-webhook-to-not-create-items-for-mega-gam-campaigns


app/controllers/webhooks_controller.rb
```diff
@@ -160,6 +160,9 @@ def handle_payment_intent_succeeded(
             email: payment_intent.purchaser.email
           }
         )
+      elsif payment_intent.campaign.present? && payment_intent.campaign.mega_gam?
+        payment_intent.successful = true
+        payment_intent.save!
```
chrisng93, line 165: i realize we actually may want to send some email here but i'm not sure what we want to do here. should we be creating a new EmailManager?

app/models/campaign.rb
```diff
@@ -102,6 +102,10 @@ def seller_distributor_pairs
     pairs
   end
 
+  def mega_gam?
+    project.present?
+  end
```
justintmckibben, line 107: mind adding a test for this in the campaign spec?

app/controllers/webhooks_controller.rb
justintmckibben: Yeah we probably should be sending an email

justintmckibben: It's also not super clear to me where the payment intent is set to successful in the other cases. I wonder if we can move all of the setting of successful out to the webhook, or move this in to a separate handler? Otherwise it looks like from here that the only case where the intent is set to successful is in the mega gam case, when that's not indeed true

justintmckibben (approved):

spec/services/webhook_manager/donation_creator_spec.rb
```diff
@@ -17,35 +17,6 @@
     expect(item.item_type).to eq('donation')
     expect(donation.item).to eq(item)
     expect(donation.amount).to eq(payload[:amount])
-    expect(PaymentIntent.last.successful).to be true
```
chrisng93, line 20: we don't need these tests anymore since they're handled by the webhook controller. in the logic, if a payment intent is successful we raise an error

justintmckibben (approved):
ty!!

PR #291, opened 2020-11-16
Don't return refunded payment intents for project

justintmckibben:


jacoblb64 (approved):
Looks good! Where are we setting refunded by the way?

justintmckibben:
> Looks good! Where are we setting refunded by the way?

inside of the webhook!

PR #292, opened 2020-11-16
Place project id on payment intent when appropriate

jacoblb64:


justintmckibben (approved):
non-blocking but a test for a valid project id and an invalid project id would be a good follow

jacoblb64:
@justintmckibben 
> non-blocking but a test for a valid project id and an invalid project id would be a good follow

We already have some tests for those use cases in `spec/requests/charges_spec.rb`, but we could also verify on PaymentIntent -- I'll merge this now and follow up later

PR #296, opened 2020-11-18
Add seller/dist to the seller_distributor_pairs array of the Campaigns API

justintmckibben:


app/models/campaign.rb
```diff
@@ -81,38 +81,27 @@ def last_contribution
...
-      pair = {
-        'distributor_id' => distributor.id,
-        'distributor_image_url' => distributor.image_url,
-        'distributor_name' => distributor.name,
-        'seller_id' => seller.id,
-        'seller_image_url' => seller.hero_image_url,
-        'seller_name' => seller.name
-      }
-      pairs.append(pair)
+      pairs.append(seller_distributor_pair(
```
aliceluo2048, line 101: Can we use `<<` instead of ``.append``?

chrisng93 (approved):

aliceluo2048 (approved):

PR #299, opened 2020-11-20
Fix seller_id in campaign api response

aliceluo2048:
GAM campaigns frontend is broken when trying at `getSeller` because we made `seller_id` an int (seller.id) instead of the string (seller_id) in the Campaigns api response in this PR: https://github.com/sendchinatownlove/ruby/pull/286

### DIFF
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -128,6 +128,7 @@ class CampaignsController < ApplicationController
     ret['amount_raised'] = campaign.amount_raised
     ret['amount_allocated'] = campaign.amount_allocated
     ret['last_contribution'] = campaign.last_contribution
+    ret['seller_id'] = campaign.seller.nil? ? "" : campaign.seller.seller_id
     ret['seller_distributor_pairs'] = campaign.seller_distributor_pairs
     ret
   end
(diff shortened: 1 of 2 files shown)


app/controllers/campaigns_controller.rb
```diff
@@ -128,6 +128,7 @@ def campaign_json(campaign: @campaign)
     ret['amount_raised'] = campaign.amount_raised
     ret['amount_allocated'] = campaign.amount_allocated
     ret['last_contribution'] = campaign.last_contribution
+    ret['seller_id'] = campaign.seller.nil? ? "" : campaign.seller.seller_id
```
justintmckibben, line 131: I think we want to update the frontend to use `seller_distributor_pairs` instead

justintmckibben (changes requested):

app/controllers/campaigns_controller.rb
aliceluo2048: Yes because there could be a project vs a seller id (normal gam vs mega gam) in the frontend- are you saying we should just keep the backend as it is?

justintmckibben: Right I've updated the regular campaigns to return the seller/dist that's related to the campaign directly as a seller_distributor_pair since we're going to be removing campaign.seller and campaign.dist in favor of seller_dist_pairs

chrisng93: i had created a ticket for this if someone wanted to take it up:
https://trello.com/c/wMetd5qi/708-megagam-migrate-campaign-seller-distributor-fields-to-use-the-campaignssellersdistributors-join-table

chrisng93: ticket is to migrate seller_id and distributor_id to the seller/distributor table and remove old code

PR #300, opened 2020-11-20
Add campaign.display_name computed field

chrisng93:
Add campaign.display_name computed field. Project name if project is associated with campaign, seller/distributor names otherwise

### DIFF
--- a/app/models/campaign.rb
+++ b/app/models/campaign.rb
@@ -107,17 +107,18 @@ class Campaign < ApplicationRecord
     pairs
   end
 
-  def name
+  def display_name
     # Right now, we require campaigns to have a project or seller assocated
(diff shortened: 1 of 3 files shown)


app/models/campaign.rb
```diff
@@ -107,6 +107,21 @@ def seller_distributor_pairs
     pairs
   end
 
+  def display_name
```
chrisng93, line 110: this is a bit of a rendering concern, so i could see a case for having the client pull the project name and not have this on the campaign model. but i think it would also be good to centralize the logic

justintmckibben: I think this is a good solution

justintmckibben (approved):

jacoblb64 (approved):

PR #302, opened 2020-12-01
Fix charges_controller amount validation to use total

Kenworthe:
This branch changes the 'amount' validation in `charges_controller#validate` to look at total amount instead of each line_item amount when validating for minimum amount.

Clears blocker to transaction fee frontend branch.


Jeremy-Xue (approved):
lgtm!

PR #303, opened 2020-12-02
transaction fee flat cost to int

Jeremy-Xue:


justintmckibben (approved):
ty for the quick turnaround!

PR #304, opened 2020-12-03
Wilson/past campaign pagination

wilsonj806:
Update `GET /campaigns` so it accepts query params and so it paginates the response. By default it returns all active campaigns, but you can add `?inactive=true` to the endpoint to fetch all inactive campaigns.

And then you can request paginated data through the standard Pagy query params 

image

NOTE: Updated the campaigns test/ test rigs and am not sure if it's super wonky or not

### DIFF
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -17,9 +17,11 @@ class CampaignsController < ApplicationController
 
       json_response(@records)
     else
-      @campaigns = valid_campaigns.order(:end_date).all
+      query = valid_campaigns.order(:end_date).all
 
-      json_response(campaigns_json)
+      @pagy, @records = pagy(query)
+
+      json_response(@records)
     end
   end
 
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -1,12 +1,23 @@
 # frozen_string_literal: true
+include Pagy::Backend
 
 class CampaignsController < ApplicationController
   before_action :set_campaign, only: %i[show update]
+  after_action { pagy_headers_merge(@pagy) if @pagy }
 
   # GET /campaigns
   def index
     @campaigns = valid_campaigns.order(:end_date).all
     json_response(campaigns_json)
+  end
+
+  # GET /campaigns/inactive
+  def index_past_campaigns
+    query = past_campaigns.order(:end_date).all
+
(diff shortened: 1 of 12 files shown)


app/controllers/campaigns_controller.rb
```diff
@@ -1,14 +1,25 @@
...
   before_action :set_campaign, only: %i[show update]
+  after_action { pagy_headers_merge(@pagy) if @pagy }
 
   # GET /campaigns
   def index
     @campaigns = valid_campaigns.order(:end_date).all
     json_response(campaigns_json)
   end
 
+  # GET /campaigns/inactive
```
justintmckibben, line 14: I wonder if we should be using query params here instead to filter so that we can query out inactive campaigns, but also filter out active campaigns. Our regular GET endpoint shouldn't return them all

wilsonj806: Yeah that probably makes more sense than building an entire new route. So if I'm updating it it'd be like:
```rb
# Get campaigns
def index
  if query_param[:inactive]
  # do whatever
  json(response_with_pagination)
  else
  # do the other thing
  json(response_without_pagination)
  end
end
```

app/controllers/campaigns_controller.rb
```diff
@@ -1,12 +1,26 @@
...
+  after_action { pagy_headers_merge(@pagy) if @pagy }
 
   # GET /campaigns
   def index
-    @campaigns = valid_campaigns.order(:end_date).all
-    json_response(campaigns_json)
+    # @NOTE(wilson) check to see if querying for inactive campaigns
+    # otherwise return active campaigns by default
+    inactive = params[:inactive]
+    if inactive == 'true'
```
jacoblb64, line 13: worried about boolean / string logic here -- should we try to convert then check it as a boolean to avoid potential issues? Not sure how we parse url query params

jacoblb64 (approved):
Looks all good to me! Left one small comments. Does this require any change on the frontend?

wilsonj806:
@jacoblb64 Yeah there's some front end changes, mostly to the GAM page/ related components

app/controllers/campaigns_controller.rb
wilsonj806: I can double check that, but yeah it's probably better to convert to a boolean. I'll fix that before marking the PR up for review

PR #305, opened 2020-12-08
Add WorkOS integration for Magic Link and Google

rjvani:


app/controllers/authorization_controller.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class AuthorizationController < ApplicationController
```
justintmckibben, line 3: we might want to test this

justintmckibben (approved):

jacoblb64 (approved):
This looks really dope man! Agreed with justin that we should have some tests / examples.

Is this a breaking change / what do we have to do to utilize this?

app/controllers/authorization_controller.rb
rjvani: Added some tests. There's a stub for the Magic Link email, I'll add that in a subsequent PR.

rjvani:
> This looks really dope man! Agreed with justin that we should have some tests / examples.
> 
> Is this a breaking change / what do we have to do to utilize this?

This isn't a breaking change. I've configured us with an API Key, account, and redirect URIs on our Google app. We just need to set up some environment variables on Heroku, and we should be good moving forward.

jacoblb64 (approved):

PR #307, opened 2020-12-11
removed is_distribution and updated campaign_id fields

billy-yuan:
I removed the `is_distribution` field and updated the `campaign_id` fields.

Would like a review to make sure my changes below align with the current business logic.

Lines refer to my diff:
- Line 111: If a campaign but a project doesnt exist, then process as a regular gam
- Lines 157-158: If a project exists, then square location id will be equal to the location id of the project

Once these changes are good to go, the front end mega gam checkout should be ready to be pushed as well since I've tested a mega gam payment and see my test transactions show up in the payment_intents table with a project ID and campaign ID. More than happy to do more testing though.

### DIFF
--- a/app/controllers/charges_controller.rb
+++ b/app/controllers/charges_controller.rb
@@ -19,7 +19,8 @@ class ChargesController < ApplicationController
     validate(
       seller_id: seller_id,
       project_id: project_id,
-      line_items: line_items
+      line_items: line_items,
+      is_distribution: charge_params[:is_distribution]
     )
 
     # Validate each Item and get all ItemTypes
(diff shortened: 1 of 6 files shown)


app/controllers/charges_controller.rb
```diff
@@ -125,12 +122,12 @@ def validate(seller_id:, project_id:, line_items:, is_distribution:)
 
                     campaign
                   elsif charge_params[:campaign_id].present?
-                    Campaign.find_by(campaign_id: campaign_id)
+                    Campaign.find_by(id: charge_params[:campaign_id])
```
justintmckibben, line 125: This can be `Campaign.find(charge_params[:campaign_id])`

app/controllers/charges_controller.rb
```diff
@@ -111,9 +108,9 @@ def validate(seller_id:, project_id:, line_items:, is_distribution:)
         raise InvalidLineItem, 'line_item.quantity must be an Integer'
       end
 
-      @campaign = if is_distribution.present?
-                    # TODO(justintmckibben): Delete this case when we start using campaign_id
-                    #                        in the frontend
+      @campaign = if charge_params[:campaign_id].present? && !project_id.present?
```
justintmckibben, line 111: I think we can replace this whole line with:
`@campaign = Campaign.find_by(id: charge_params[:campaign_id]) if charge_params[:campaign_id].present?`

justintmckibben (changes requested):
I think some tests need to be updated too!

stzhng:
as suggested by previous standup this pr is blocked because the secrets or environment variable of location id is not defined. 

1) you can set it on your fork of the repo in github/secrets and assign random numbers
2) straight forwardsolution is closing this PR and adding sendchinatownlove/ruby as an upstream master and pushing to a branch. @billy-yuan 


image

PR #308, opened 2020-12-12
test rake tests

billy-yuan:


billy-yuan:
This PR is exactly the same as master but i just added a comment in one line, yet tests are failing. What would the root cause be?

PR #309, opened 2020-12-13
Add neighborhood to campaign in favor of location

justintmckibben:
Campaigns no longer show a full address, so a location
relationship was overkill. We wanted to make the API and
model simplier so are replacing location with just a raw
string.


justintmckibben:
We need to update the frontend to use this over the location after this merges

rjvani (approved):
Should `campaigns.location_id` in `schema.rb` have updated and removed `null: false`?

PR #311, opened 2020-12-14
Add Apex handling in WebhookValidator and PaymentCreator

slants0:


app/services/square_manager/webhook_validator.rb
```diff
@@ -34,9 +34,17 @@ def call
         )
       )
 
+      apex_signature = Base64.strict_encode64(
+        OpenSSL::HMAC.digest(
+          'sha1',
+          ENV['APEX_WEBHOOK_SIGNATURE_KEY'],
+          string_to_sign
+        )
+      )
```
justintmckibben, line 43: Since this is the third time we're using this, can we abstract this out into a private method?

app/services/square_manager/webhook_validator.rb
```diff
@@ -34,9 +34,17 @@ def call
...
+
       # Hash the signatures a second time (to protect against timing attacks)
       # and compare them
-      unless [Digest::SHA1.base64digest(string_signature), Digest::SHA1.base64digest(think_chinatown_signature)].include?(Digest::SHA1.base64digest(callback_signature))
+      unless [Digest::SHA1.base64digest(string_signature), Digest::SHA1.base64digest(think_chinatown_signature), Digest::SHA1.base64digest(apex_signature)].include?(Digest::SHA1.base64digest(callback_signature))
```
justintmckibben, line 47: same might also be worth abstracting out `Digest::SHA1.base64digest(blah)`

justintmckibben (changes requested):

app/services/square_manager/webhook_validator.rb
slants0: Ofc

app/services/square_manager/webhook_validator.rb
slants0: Ditto

jacoblb64 (approved):

PR #314, opened 2020-12-15
add non-profit's fee_id to payment intent

billy-yuan:
Updated the `charges_controller` so that the `fee_id` from a non-profit is included in a `payment_intent`.

Logic I used to determine the `fee_id` is:
- Is there a campaign?
- If so, is there a non-profit associated with that campaign?
- If so, is there a `fee_id` associated with that non-profit?
- If so, return the `fee_id`

If any of the questions above is "no", then don't return anything

### DIFF
--- a/app/controllers/charges_controller.rb
+++ b/app/controllers/charges_controller.rb
@@ -195,8 +195,8 @@ class ChargesController < ApplicationController
     # Creates a pending PaymentIntent. See webhooks_controller to see what
     # happens when the PaymentIntent is successful.
     
-    # Check for whether a campaign exits. If a campaign exists, check if the campaign has a non-profit associated with it. 
-    # If a non-profit exists, check if the non-profit has a fee ID. If so, the set fee ID equal to fee_id
+    # Check for whether a campaign exits. If a campaign exists, then check whether the campaign has a non-profit associated with it. 
+    # If a non-profit exists, then check whether the non-profit has a fee ID. If so, then set fee ID equal to fee_id
     fee_id = if @campaign && @campaign[:nonprofit_id] && Nonprofit.find_by(id: @campaign[:nonprofit_id])[:fee_id]
               Fee.find_by(id: Nonprofit.find_by(id: @campaign[:nonprofit_id])[:fee_id])[:id]
              end
--- a/app/controllers/charges_controller.rb
+++ b/app/controllers/charges_controller.rb
@@ -194,6 +194,13 @@ class ChargesController < ApplicationController
 
     # Creates a pending PaymentIntent. See webhooks_controller to see what
     # happens when the PaymentIntent is successful.
+    
+    # Check for whether a campaign exits. If a campaign exists, check if the campaign has a non-profit associated with it. 
+    # If a non-profit exists, check if the non-profit has a fee ID. If so, the set fee ID equal to fee_id
+    fee_id = if @campaign && @campaign[:nonprofit_id] && Nonprofit.find_by(id: @campaign[:nonprofit_id])[:fee_id]
+              Fee.find_by(id: Nonprofit.find_by(id: @campaign[:nonprofit_id])[:fee_id])[:id]
+             end
+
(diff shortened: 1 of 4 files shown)


app/controllers/charges_controller.rb
```diff
@@ -194,6 +194,13 @@ def create_square_payment_request(
 
     # Creates a pending PaymentIntent. See webhooks_controller to see what
     # happens when the PaymentIntent is successful.
+    
+    # Check for whether a campaign exits. If a campaign exists, then check whether the campaign has a non-profit associated with it.
```
justintmckibben, line 198: nit: Mind keeping the line length under 120? (Whatever rubocop is configured to)

app/controllers/charges_controller.rb
```diff
@@ -194,6 +194,13 @@ def create_square_payment_request(
 
     # Creates a pending PaymentIntent. See webhooks_controller to see what
     # happens when the PaymentIntent is successful.
+    
+    # Check for whether a campaign exits. If a campaign exists, then check whether the campaign has a non-profit associated with it. 
+    # If a non-profit exists, then check whether the non-profit has a fee ID. If so, then set fee ID equal to fee_id
+    fee_id = if @campaign && @campaign[:nonprofit_id] && Nonprofit.find_by(id: @campaign[:nonprofit_id])[:fee_id]
+              Fee.find_by(id: Nonprofit.find_by(id: @campaign[:nonprofit_id])[:fee_id])[:id]
```
justintmckibben, line 201: nit: This can be `Fee.find(Nonprofit.find(@campaign([:nonprofit_id][:fee_id][:id])`

app/controllers/charges_controller.rb
```diff
@@ -194,6 +194,13 @@ def create_square_payment_request(
 
     # Creates a pending PaymentIntent. See webhooks_controller to see what
     # happens when the PaymentIntent is successful.
+    
+    # Check for whether a campaign exits. If a campaign exists, then check whether the campaign has a non-profit associated with it. 
+    # If a non-profit exists, then check whether the non-profit has a fee ID. If so, then set fee ID equal to fee_id
+    fee_id = if @campaign && @campaign[:nonprofit_id] && Nonprofit.find_by(id: @campaign[:nonprofit_id])[:fee_id]
```
justintmckibben, line 200: If you don't mind, do you mind writing a test for this in the charges_controller_spec?

app/controllers/charges_controller.rb
```diff
@@ -203,7 +210,8 @@ def create_square_payment_request(
       recipient: recipient,
       campaign: @campaign,
       metadata: metadata,
-      project: Project.find_by(id: project_id)
+      project: Project.find_by(id: project_id),
```
justintmckibben, line 213: nit: Project.find(...)

justintmckibben: What happens if there is no `project_id`?

justintmckibben: The caveat of find is that I think that is crashes if it doesn't exist, but find_by returns nil

app/controllers/charges_controller.rb
justintmckibben, line 200: This is complex enough to where I'd abstract this into a private method called `_get_fee` or something

justintmckibben (approved):
Nothing blocking! Just some nits. The test would be amazing though!

app/controllers/charges_controller.rb
billy-yuan: I didn't add this line but if a transaction has no project (e.g. single-seller GAM, donation), there wouldn't be a project so nil should be passed otherwise the transaction won't complete unless I'm mistaken

For future multi-seller GAMs that do not have a project, we are passing the campaign, which should be enough to create a charge?

billy-yuan:
i made some changes based on justin's feedback but there seems to have been an issue with merging, so i'm going to push the changes to a new branch

PR #315, opened 2020-12-16
[Bug] Fix transaction_fee in webhook_controller

Kenworthe:
- Should now properly exclude transaction_fee line items when iterating in webhooks controller
- Should fix email double-send.

### DIFF
--- a/app/controllers/webhooks_controller.rb
+++ b/app/controllers/webhooks_controller.rb
@@ -114,10 +114,11 @@ class WebhooksController < ApplicationController
     end
 
     items = JSON.parse(payment_intent.line_items)
-    recipient = payment_intent.recipient
     is_donation = false
     save_payment_intent = false
-    items = items.reject { |item| item['type'] == 'fee' }
+
+    items.reject! { |item| item['item_type'] == 'transaction_fee' }
+
     items.each do |item_json|
       # TODO(jtmckibb): Add some tracking that tracks if it breaks somewhere
       # here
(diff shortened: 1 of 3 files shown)


app/controllers/webhooks_controller.rb
```diff
@@ -114,10 +114,11 @@ def handle_payment_intent_succeeded(
     end
 
     items = JSON.parse(payment_intent.line_items)
-    recipient = payment_intent.recipient
```
Kenworthe, line 117: unused var

app/controllers/webhooks_controller.rb
```diff
@@ -114,10 +114,11 @@ def handle_payment_intent_succeeded(
...
     is_donation = false
     save_payment_intent = false
-    items = items.reject { |item| item['type'] == 'fee' }
+
+    items.reject! { |item| item['item_type'] == 'transaction_fee' }
```
justintmckibben, line 120: You can merge this, but would be great to do a follow up that includes a test with a `transaction_fee` line item

justintmckibben (approved):

app/controllers/webhooks_controller.rb
justintmckibben: and in that test could you also add a test inside of the payment_intent_spec to test the amount on the intent?

Kenworthe: yup. done! re-review?

justintmckibben: Oh perf! yeah there is another test in the payment_intent_spec that tests the method "amount" to make sure that the amount of the payment intent is calculated properly, and I wanted to make sure that adding this new item_type doesn't break that

PR #317, opened 2020-12-19
add non-profit's fee_id to payment intent

billy-yuan:
There was a rebase issue in my previous PR, so I'm just incorporating comments from the [previous PR](https://github.com/sendchinatownlove/ruby/pull/314):

**This PR will fix the following issues in the `charges_controller`:**
1. Lines 206-220: **The `fee_id` of nonprofits will now be passed when a payment intent is created** (same change as the previous PR above). Test has also been created for this. 
2. Line 128,133, and 158: **Payment intents for campaigns that don't have a seller** (i.e. multi-seller GAM) **can now be created.**

**Tests created**
1.  When a campaign has a nonprofit with a fee_id, it should create a payment_intent that has the campaign_id and fee_id
2. When a campaign has a project_id and a nonprofit with a fee_id, it should create a payment_intent that has the project_id, campaign_id and fee_id

**Notes**
1. The chunk of this PR is from the `charges_controller_spec`. I had to refactor the code a bit so that I could add my 2 tests.
2. Removing `is_distribution` should be done in a separate PR since some tests may have to be edited, and the logic in the charges controller will also have to be updated.

### DIFF
--- a/spec/controllers/charges_controller_spec.rb
+++ b/spec/controllers/charges_controller_spec.rb
@@ -1,3 +1,4 @@
+
 # frozen_string_literal: true
 
 require 'rails_helper'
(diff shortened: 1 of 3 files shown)


spec/controllers/charges_controller_spec.rb
```diff
@@ -28,53 +29,195 @@
...
+      'amount' => 50, 
+      'currency' => 'usd', 
+      'quantity' => 1, 
+      'item_type' => 'donation' 
+    }
+  end
+
+  # TODO (billy-yuan): Update create_line_items so that transaction fees can be
+  # included as a separate line_items group
+  def create_line_items(extra_line_items = {})
```
billy-yuan, line 176: Examples of `extra_line_items` include `project_id` and `seller_id` but there's no typechecking or anything in place at the moment

PR #321, opened 2021-01-16
added callback to payment intent to make campaign inactive if goal is met

billy-yuan:
Added a callback to the `PaymentIntent` model that makes the campaign inactive if goal is met after a payment intent is created.

I haven't checked to see whether this works for normal campaigns since the logic is using the PaymentIntent model.

Also open to using a different approach for checking whether a campaign has met its goal

### DIFF
--- a/app/models/payment_intent.rb
+++ b/app/models/payment_intent.rb
@@ -44,10 +44,23 @@ class PaymentIntent < ApplicationRecord
   belongs_to :campaign, optional: true
   belongs_to :project, optional: true
 
+  after_create :make_campaign_inactive_if_met_goal
+
   def amount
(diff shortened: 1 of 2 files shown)


justintmckibben:
Messaged you personally, but I am worried that this type of approach could lead to really hard to follow dependencies in the future. I could see this becoming wiry if we continue to follow this pattern, and we'd find ourselves in a position where we'd create a PaymentIntent, and for some reason all our Passport models are being affected and we can't figure out why.

I think clear definition of dependencies using dependency injection might be the move here, or moving the logic into the webhooks_controller. Happy to leave this for now and refactoring in the future if it becomes hard to follow.

Thanks for doing this!

PR #322, opened 2021-01-18
Uploading Receipts

jacoblb64:
Front end MR: https://github.com/sendchinatownlove/sendchinatownlove.github.io/pull/474

**Background:**
- We want to store receipts in Google Cloud Storage (GCS)
- We want to have them uploaded _directly_ from the front-end, to avoid strain on the back-end
- The front-end cannot store private env variables (nor would it be probably be within best practices anyway in this case)
- Our back-end has private env variables with Heroku
- GCS requires auth, and has a few different methods, the most fitting in this case is probably to use a [Service Account](https://cloud.google.com/storage/docs/getting-service-account)

**Basic Overview:**
- The backend will contain the Service Account credentials
- These are expected to be in a json file, but we can use [this common Heroku Buildpack](https://github.com/gerywahyunugraha/heroku-google-application-credentials-buildpack) to build this from an env variable
- The front-end will ask the backend for a signed url for a given filename, and the backend will authenticate and send a [Signed URL](https://cloud.google.com/storage/docs/access-control/signed-urls) with an expiration
- The frontend will receive the url and upload the receipt without authentication
- GCS has a [CORS policy](https://cloud.google.com/storage/docs/configuring-cors) to restrict this action to our sites (configuration done directly)

I think this is the best way to achieve what we want securely, and following best practices.


justintmckibben (approved):
A lot of magic going on here, but I trust it works!

spec/requests/gcs_request_spec.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Gcs', type: :request do
+  describe 'GET /create' do
```
stanzheng, line 6: should this be a PUT for ccreate?

app/controllers/gcs_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  # POST /gcs
+  def create
+
+    bucket_name = 'scl-lny-receipts'
+    file_name = gcs_params[:file_name]
+    file_type = gcs_params[:file_type]
+
+    storage = Google::Cloud::Storage.new
+    storage_expiry_time = 15 * 60 # 5 minutes
```
stanzheng, line 15: i think this is actually 15 minutes but shouldn't matter too much. if someone has the client open for a long period of time they might get an error but i dont know if sessions will last > 15 minutes

app/controllers/gcs_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+require 'google/cloud/storage'
+
+class GcsController < ApplicationController
+
+  # POST /gcs
+  def create
+
+    bucket_name = 'scl-lny-receipts'
+    file_name = gcs_params[:file_name]
```
stanzheng, line 11: since this is a shared bucket and it looks like the client is updating the filename? if theres a convention i recommend storing possibly a DATESTAMP_VENDOR_NAME or something of the like to avoid conflicts. 

Also you can use possibly folders/prefixes but this should be ok since if we expect < 5000 receipts it should be managable

spec/requests/gcs_request_spec.rb
jacoblb64: Ah it's supposed to be a `POST /gcs` - this is just the auto-generated tests, and I messed up the `rails g` syntax, I'll update appropriately.

app/controllers/gcs_controller.rb
jacoblb64: Good catch, I forgot to update my comment. The UI flow has the getting signed URL and upload back-to-back, so these should in theory never be far apart, I think we could lower this back to 5, but I'll do some testing.

app/controllers/gcs_controller.rb
jacoblb64: My idea was to have the front-end construct the filename as it has the relevant information (user email address, vendor name, etc.). I think with our scale all in the same bucket without files is probably fine for now, but we can update later if need be. In the front-end MR I did `[email before @ sign]-[full ISO timestamp].[ext]`, but maybe `[full-escaped-email]-[vendor name]-[full ISO timstamp].ext` will actually be better. What do you think?

jacoblb64:
@stanzheng and @justintmckibben take a look again, marked ready for review. I added a unit test that does actually call GCS, but as that's really more of an integration test, I've set it to skip unless we want to run it manually.

Also, I think I should install the heroku buildpack before merging this, let me know if you're comfortable with me going ahead and doing that.

PR #323, opened 2021-01-19
Mega gam allocation endpoint

chrisng93:
Creates mega gam campaign allocation endpoint. It takes in a campaign id (in url) and seller/distributor/amount (in body) and creates GiftCardDetails/GiftCardAmount/Item

The PR is still missing some tests

[Ticket](https://trello.com/c/N1fuggcX/696-megagam-create-allocation-endpoint-that-takes-in-a-campaign-id-and-list-of-sellers-amounts-and-creates-items-for-the-mega-gam-ca)


app/controllers/campaigns_controller.rb
```diff
@@ -50,6 +50,30 @@ def associate_seller_distributor
...
+  # POST /campaigns/:id/gift_card
+  def generate_campaign_gift_cards
+    request_params = generate_campaign_gift_cards_params
+    unless @campaign.amount_raised > @campaign.amount_allocated + request_params['gift_card_amount_cents']
+      raise InvalidLineItem, "Request amount exceeds unallocated amount in campaign. Unallocated amount: #{
+        @campaign.amount_raised - @campaign.amount_allocated
+      }"
+    end
+
+    WebhookManager::GiftCardCreator.call(
```
chrisng93, line 62: not sure if we wanted to send out receipts for this as well?

app/controllers/campaigns_controller.rb
```diff
@@ -105,6 +129,24 @@ def associate_seller_distributor_params
     set_distributor
     @seller = Seller.find(params[:seller_id])
 
+    ret
+  end
+
+  def generate_campaign_gift_cards_params
+    params.require(:distributor_id)
+    params.require(:gift_card_amount_cents)
```
justintmckibben, line 137: I personally think the "cents" should be implied for anything on our API

app/controllers/campaigns_controller.rb
```diff
@@ -50,6 +50,30 @@ def associate_seller_distributor
     json_response(campaign_json)
   end
 
+  # POST /campaigns/:id/gift_card
```
justintmckibben, line 53: So the idea is that they'd hit this over and over? A bulk/battch version of this endpoint might be a good follow

app/controllers/campaigns_controller.rb
```diff
@@ -50,6 +50,30 @@ def associate_seller_distributor
...
+        amount: request_params['gift_card_amount_cents'],
+        single_use: true,
+        project_id: @campaign.project_id,
+        campaign_id: @campaign.id,
+        distributor_id: @distributor.id,
+      }
+    )
+
+    json_response({
+      'amount_left_to_allocate': @campaign.amount_raised - @campaign.amount_allocated
```
justintmckibben, line 73: Can we add this number to the Campaign GET

app/controllers/campaigns_controller.rb
justintmckibben: I don't think we wanna send out receipts

app/services/webhook_manager/gift_card_creator.rb
```diff
@@ -3,30 +3,37 @@
...
+                                                  campaign_id: campaign_id,
                                                 })
 
         gift_card_detail = GiftCardDetail.create!(
           expiration: Date.today + 1.year,
           item: item,
           gift_card_id: GiftCardIdGenerator.generate_gift_card_id,
           seller_gift_card_id: GiftCardIdGenerator.generate_seller_gift_card_id(seller_id: seller_id),
-          recipient: payment_intent.recipient,
+          recipient: payment_intent ? payment_intent.recipient : distributor_contact,
```
justintmckibben, line 36: I think we should just pass in recipient so that we don't have to pass in the Campaign in the first place

justintmckibben (approved):
LGTM just some nits

app/models/item.rb
```diff
@@ -33,7 +33,7 @@
 class Item < ApplicationRecord
   belongs_to :campaign, optional: true
   belongs_to :payment_intent, optional: true
-  belongs_to :purchaser, class_name: 'Contact'
+  belongs_to :purchaser, class_name: 'Contact', optional: true
 
   # Must have seller or project, but not both
```
chrisng93, line 38: so one thing is that the items that are created by the allocation endpoint have a seller and project. we use the project as a holder for the campaign money, but the item needs a specific seller associated as well. does anyone happen to remember why we have an xor here? would it break anything if i removed this validation?

justintmckibben: It won't break anything! We had it because at the time of creation it only knows one or the other

app/controllers/campaigns_controller.rb
chrisng93: yeah i changed this endpoint to make it bulk. i had initially had that in the ticket, but i was stuck on some of the nested validation in rails for a bit

app/services/webhook_manager/gift_card_creator.rb
chrisng93: this pr ballooned a bit since i had to change a bunch of different assumptions. i'd prefer to do this in a follow up

app/controllers/campaigns_controller.rb
chrisng93: got it

app/controllers/campaigns_controller.rb
chrisng93: changed this to be "gift_card_amount"

app/controllers/campaigns_controller.rb
chrisng93: sure i can do that in a follow up

app/models/item.rb
chrisng93: ok got it, thanks!

chrisng93:
@justintmckibben i addressed your comments / added a bit of logic here. would appreciate another look at it when you have a chance

chrisng93:
@justintmckibben bump on this - added a bunch of logic here and it'd be good to get a set of eyes on it

app/controllers/campaigns_controller.rb
```diff
@@ -50,6 +50,39 @@ def associate_seller_distributor
...
+          single_use: true,
+          project_id: @campaign.project_id,
+          campaign_id: @campaign.id,
+          distributor_id: gift_card['distributor_id'],
+          seller_id: gift_card['seller_id'],
+        }
+      )
+    end
+
+    json_response({'unallocated_amount': @campaign.amount_unallocated})
```
jacoblb64, line 83: Got a bunch of real nitpicky comments you can take or leave, just thought I'd put them all in one comment 😂 

- We usually do the permitting / requiring in a separate function to be called, also I feel like you might want to explicitely not just permit, but **require** the `gift_cards` param here
- You can add your function to the allow-list of `before_actions` and `set_campaign` will be invoked automatically for you
- If you wanna be real fancy you can do a `reduce` for your sum, totally optional though!
- I'm not sure if rails caches the value for `@campaign.amount_unallocated` -- that calls some underlying DB queries, so I'd want to avoid calling it multiple times per API hit if possible. If it does, I'd recommend saving this to a local variable to save time/memory.
- I think `InvalidParameter` is probably a better exception, I think that one was writteexplicitlyly for `line_items` on the `/charges` endpoint... but maybe we're not using it that way anymore anyway, and they take the same path in `exception_handler.rb`
- Maybe return status code `:created` to let the caller know everything went well? And that this is just the remainder left?

jacoblb64 (approved):
Looks good to me man, lots of tricky logic here that I think you've handled well. Left you a plethora of nitpicky comments if you're feeling up for it, but none of them are blocking to me.

I'd say ping @justintmckibben one last time on slack and get him to sign off though 😂

app/models/item.rb
jacoblb64: Yeah, feel free to change this if things have changed! This was an initial assumption and we put these guards in as a safety net!

app/controllers/campaigns_controller.rb
chrisng93: thanks for the suggestions. i updated the code to reflect them and pinged @justintmckibben about a re-stamp

justintmckibben (approved):
Looks great! I do think it would be good to write a test in the campaigns_spec, at least so that people understand how to use the new endpoint

app/controllers/campaigns_controller.rb
jacoblb64: Awesome man, looks really good now, hope these were helpful!

chrisng93:
@justintmckibben added some tests!

PR #325, opened 2021-01-25
Implement custom payment intents

slants0:
## Issue
We're receiving cash donations that circumvent our usual Square payment flow. As such, we need ways to create donations with payment intents that are not from Square.

## Solution
Adding an `origin` column to the payment intents table to separate Square and non-Square payment intents. Thinking here is they are technically the same type of object but with slight variations and should belong in the same table in the database. Also decided to incorporate the changes into the existing payment intents model instead of creating a new one as it seems behavior will largely be the same for both. I'm not married to this though and perhaps inheriting the existing payment intents model into a `CustomPaymentIntents` model might be better in the long run depending on what actions custom payments need to perform.


jacoblb64 (approved):
Looks good to me! I'm wondering if we should add more detail than just "custom", or at at least add another column for details. I'm thinking if we'd want to track the type, e.g. "cash" or "check", and if so then possibly a place to capture the check number or other details. @justintmckibben, thoughts?

app/models/payment_intent.rb
```diff
@@ -36,8 +36,13 @@
...
 #
 class PaymentIntent < ApplicationRecord
-  validates_presence_of :square_payment_id, :square_location_id
-  validates_uniqueness_of :square_payment_id, allow_nil: false
+  ORIGIN_TYPES = [
+    CUSTOM = 'custom'.freeze,
+    SQUARE = 'square'.freeze
+  ].freeze
+
+  validates_presence_of :square_payment_id, :square_location_id, if: -> { origin == SQUARE }
```
justintmckibben, line 44: Nice! Do you mind adding a test for this?

app/models/payment_intent.rb
```diff
@@ -36,8 +36,13 @@
...
+  validates_uniqueness_of :square_payment_id, allow_nil: true
```
justintmckibben, line 45: Can we verify that this is true in prod before merging?

db/migrate/20210124231324_add_non_square_payments_to_payment_intents.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddNonSquarePaymentsToPaymentIntents < ActiveRecord::Migration[6.0]
```
justintmckibben, line 3: Running this migration should update schema.rb

justintmckibben (changes requested):

app/models/payment_intent.rb
slants0: @justintmckibben I'm not sure I follow - verifying all `square_payment_ids` are unique in the DB?

slants0:
> Looks good to me! I'm wondering if we should add more detail than just "custom", or at at least add another column for details. I'm thinking if we'd want to track the type, e.g. "cash" or "check", and if so then possibly a place to capture the check number or other details. @justintmckibben, thoughts?

I think this would be a good idea - but I'm unsure what the existing `metadata` column does, perhaps the information you're referring to could just be slotted into there?

app/models/payment_intent.rb
jacoblb64: @justintmckibben and @slants0 I went ahead and checked the Prod DB -- looks like everything is indeed unique :)

jacoblb64 (approved):
Thanks for expanding the tests! I'm cool to merge this in as-is. As you said we have ways of including other payment metadata if we so choose -- either in line items or metadata.

If you're curious, `metadata` was a column we added for LUC to send additional information from the front-end to the back-end and not to square, where line_items are sent to square for processing. So far it's only been used for the additional LUC information (lantern name, etc.)

slants0:
Thanks for the context! 👍

app/models/payment_intent.rb
slants0: Woot thanks!

PR #326, opened 2021-01-25
List all rewards

phiyan:
https://trello.com/c/BvM60tib/796-lny-crawl-create-get-endpoint-for-rewards-index

*includes database migration*


spec/models/reward_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+#  image_url   :string
+#  name        :string
+#  total_value :integer
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#
+require 'rails_helper'
+
+RSpec.describe Reward, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
justintmckibben, line 15: you might can do a "creating the thing creates the thing" here

spec/requests/rewards_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe '/rewards', type: :request do
+  describe 'GET /index' do
```
justintmckibben, line 6: you probably also wanna add a test that creating X things returns X things

justintmckibben (approved):
Looks good just some test nits!

PR #332, opened 2021-01-28
Generate LNY crawl receipts through website donations/giftcards

Jeremy-Xue:


app/services/webhook_manager/crawl_receipt_creator.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    def initialize(params)
+      @payment_intent = params[:payment_intent]
+      @amount = params[:amount]
+      @contact_id = payment_intent.purchaser_id
+      @payment_intent_id = payment_intent.id
+    end
+
+    def call
+      if Date.today.month == 2 && @amount >= 1000
```
justintmckibben, line 14: This is good enough for now, but we'd want it to be the month of February in EST. Right now it's using UTC. 

Also nits: if we could abstract these two things into constants on the top, that would be good. 

`FEBRUARY_EST = Time.now <= ... && Time.now >= ...`
`@amount >= CrawlReceipt::CRAWL_RECEIPT_MIN_AMOUNT`

I think go ahead and merge as-is though

justintmckibben (approved):

app/controllers/webhooks_controller.rb
```diff
@@ -200,6 +208,10 @@ def handle_payment_intent_succeeded(
             }
           )
           save_payment_intent = true
+          WebhookManager::CrawlReceiptCreator.call(
```
justintmckibben, line 211: Is there not one place at the bottom of the webhook that we can put this to consolidate these three calls into one?

jacoblb64: Or, if this is called in the majority, reverse the logic to only _not_ call this in the opposite case

spec/requests/webhooks_spec.rb
```diff
@@ -74,6 +76,28 @@
       create(:donation_detail, item: item, amount: 10_00)
     end
 
+    describe 'crawl receipts for donations outside february/lny crawl' do
+      let(:amount) { 5000 }
+      let(:item_type) { 'donation' }
+      let(:seller_id) { seller1.seller_id }
+
+      before do
+        Timecop.freeze(Date.new(2021, 3, 15))
```
jacoblb64, line 85: could be good to test edge cases at the beginning and end of month. I don't think we necessarily care if a few extra get in, but we do want to be careful not to reject at the wrong time -- might also make sure we cover @justintmckibben's point of UTC vs ET

jacoblb64 (approved):
Looks good! Maybe some small timezone issues as @justintmckibben pointed out, but those can be fixed later if you prefer.

app/services/webhook_manager/crawl_receipt_creator.rb
Jeremy-Xue: revising now!

Jeremy-Xue: assigning the time to a constant caused the timecop freeze to fail since it doesn't re-evaluate Time.now after freezing time in the tests, using the same value in EST though instead!

PR #336, opened 2021-01-29
Now associating receipts with their redemption

jacoblb64:
- will add tests to cover new cases by this later


Jeremy-Xue (approved):
lgtm! hopefully we can write tests checking that the crawl receipt redemptions are set after the testing party

PR #338, opened 2021-01-31
Add Active attribute to participating sellers

Jeremy-Xue:
This will allow us to only show lny crawl participating sellers, and not ones from the old crawl

### DIFF
--- a/app/controllers/participating_sellers_controller.rb
+++ b/app/controllers/participating_sellers_controller.rb
@@ -4,7 +4,7 @@ class ParticipatingSellersController < ApplicationController
   before_action :set_participating_seller, only: %i[show update]
 
   def index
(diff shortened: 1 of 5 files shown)


app/controllers/participating_sellers_controller.rb
```diff
@@ -31,7 +31,8 @@ def participating_seller_params
     params.permit(
       :name,
       :seller_id,
-      :stamp_url
+      :stamp_url,
+      :active
```
Jeremy-Xue, line 35: wrong commit message, allow creation with active flag

Jeremy-Xue: rebased to change commit ordering that was a little wonky

PR #341, opened 2021-01-31
Unredeem Receipts on Redemption Deletion

Jeremy-Xue:
In order to destroy the grand prize reward, we need to destroy the redemptions associated with the reward. However, we can't destroy a redemption with associated receipts, so this PR allows us to 'unredeem' crawl_receipts when the associated redemption is destroyed.


jacoblb64 (approved):
Don't quite fully understand the use case, but I trust you! Let's chat on slack :)

PR #342, opened 2021-02-05
Add scripts for custom payment intent and gift-cards generation

slants0:
Creates scripts for generating custom payment intents and gift cards for December GAM.


app/models/payment_intent.rb
```diff
@@ -47,7 +47,7 @@ class PaymentIntent < ApplicationRecord
   validates_uniqueness_of :square_payment_id, allow_nil: true
   has_many :items
   belongs_to :purchaser, class_name: 'Contact'
-  belongs_to :recipient, class_name: 'Contact'
+  belongs_to :recipient, class_name: 'Contact', optional: true
```
justintmckibben, line 50: If you remove this constraint, do you mind adding an assert in the GiftCardCreator that says that in order to create a gift card we need a recipient?

justintmckibben (approved):
LGTM!

spec/models/payment_intent_spec.rb
```diff
@@ -47,7 +47,6 @@
       it { should validate_uniqueness_of(:square_payment_id) }
       it { should have_many(:items) }
       it { should belong_to(:purchaser) }
-      it { should belong_to(:recipient) }
```
Jeremy-Xue, line 50: nit: I think you can use `it { should belong_to(:recipient).optional }` here!

scripts/generate_gift_cards_for_campaign.rb
```diff
@@ -0,0 +1,20 @@
+def max_campaign(payment_intent_id, campaign_id)
```
Jeremy-Xue, line 1: are we passing in the `payment_intent_id` that we generate with the `custom_payment_intent` script?

scripts/generate_gift_cards_for_campaign.rb
```diff
@@ -0,0 +1,20 @@
...
+  Item.transaction do
+    campaign = Campaign.find(campaign_id)
+    payment_intent = PaymentIntent.find(payment_intent_id)
+    payment_intent.update!(recipient: campaign.distributor.contact)
+
+    (1..(campaign.target_amount / campaign.price_per_meal)).each do |_|
```
Jeremy-Xue, line 7: do we have to take into account the amount already raised? (I'm not sure what the use case for this script is -- from the method name my understanding is that it's purpose is to max out an existing campaign?)

Jeremy-Xue (approved):
lgtm! some clarifying questions and small nits

Jeremy-Xue:
> lgtm! some clarifying questions and small nits

follow up: this maybe could be refactored into a rake task

app/services/webhook_manager/gift_card_creator.rb
```diff
@@ -16,33 +16,37 @@ def initialize(params)
...
-        gift_card_detail = GiftCardDetail.create!(
-          expiration: Date.today + 1.year,
-          item: item,
-          gift_card_id: GiftCardIdGenerator.generate_gift_card_id,
-          seller_gift_card_id: GiftCardIdGenerator.generate_seller_gift_card_id(seller_id: seller_id),
-          recipient: payment_intent ? payment_intent.recipient : distributor_contact,
-          single_use: single_use
-        )
-        GiftCardAmount.create!(value: amount, gift_card_detail: gift_card_detail)
+          gift_card_detail = GiftCardDetail.create!(
```
justintmckibben, line 38: nit: could you update these tabs to spaces?

justintmckibben: protip: `rubocop -a` will autoformat all your ruby

app/services/webhook_manager/gift_card_creator.rb
```diff
@@ -16,33 +16,37 @@ def initialize(params)
...
 
     def call
-      ActiveRecord::Base.transaction do
-        seller = Seller.find_by(seller_id: seller_id)
-        distributor = Distributor.find_by(id: distributor_id)
-        distributor_contact = nil
-        if distributor
-          distributor_contact = Contact.find_by(id: distributor.contact_id)
-        end
+      if payment_intent && payment_intent.recipient.blank?
```
justintmckibben, line 19: `if payment_intent.exists? ...`

justintmckibben (approved):

app/services/webhook_manager/gift_card_creator.rb
```diff
@@ -16,33 +16,37 @@ def initialize(params)
...
+        raise InvalidParameterError, 'Payment intent must have a recipient in order to create gift cards'
```
justintmckibben, line 20: do you mind adding a test for this?

lib/tasks/max_campaign_goals.rake
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+namespace :campaign_goals do
+  task :max, %i[payment_intent_id campaign_id] => [:environment] do |_task, args|
+    desc 'Max out campaign goals with gift cards'
+    Item.transaction do
+      campaign = Campaign.find(args.campaign_id)
+      payment_intent = PaymentIntent.find(args.payment_intent_id)
+      payment_intent.update!(campaign: campaign)
```
slants0, line 9: Calling out this line that @Jeremy-Xue and I added - the item creator needed the payment intent to have a campaign so that the individual gift cards will be associated with a campaign!

slants0:
This looks all set to me but going to wait on a restamp from @justintmckibben or @jacoblb64 before merging!

jacoblb64 (approved):
All looks good to me! Some tricky stuff in here you've handled well. I can imagine at some later point we might want to pass in more stuff to the `line_items` as arguments to the task, but this is a great foundation and we can make those changes later if/when they're needed!

PR #345, opened 2021-02-17
Add authentication validation route

wilsonj806:
Add simple session validation route. Also update development and production environment config to accept cookies in certain routes, and also allow pagination headers in the gift cards route.

Not sure how to handle testing it though. It's kind of funky


jacoblb64 (approved):
Nice and easy, looks good! A unit test would be nice, but for something this simple not a blocker

PR #346, opened 2021-02-18
Remove min amount for crawl receipt

justintmckibben:
We found that having a min amount for the food
crawl was difficult for some merchants. We
decided to remove it altogether and in order to
be consistent across the board we want to also
remove it for donations.


wilsonj806 (approved):
looks good!

jacoblb64 (approved):
Looks good!

PR #347, opened 2021-02-22
LNY crawl redemption email rake task

Jeremy-Xue:
Email sent locally:

Screen Shot 2021-02-21 at 9 32 58 PM

Copy: https://docs.google.com/document/d/1QUFaA3swXXa1XUoUAd_nIttvgFHo_FCdULwuOIYFPZ4/edit

We plan to use the rake task to retroactively send out email confirmations for the (as of 2/21) 25 contacts who have redeemed 28 entries. 

This is [followed up by a PR](https://github.com/sendchinatownlove/ruby/pull/348) that will automate this process (with changes to the email body)

### DIFF
--- a/app/services/email_manager/lny_crawl_redemption_sender.rb
+++ b/app/services/email_manager/lny_crawl_redemption_sender.rb
@@ -21,7 +21,7 @@ module EmailManager
         </p>
 
         <p>
-          • Number of Raffle Tickets Entered: #{number_of_raffle_tickets_entered}
+          &emsp;• Number of Raffle Tickets Entered: #{number_of_raffle_tickets_entered}
         </p>
       "
       end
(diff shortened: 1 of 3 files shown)


jacoblb64 (approved):
Looks all good to me! Always weird to see emojis / unicode characters in code, but if it works, it works!

As for the copy, would indenting the bullet with number of entries make it a little easier to see which basket it's referring too? Not a blocker just curious

Jeremy-Xue:
> Looks all good to me! Always weird to see emojis / unicode characters in code, but if it works, it works!
> 
> As for the copy, would indenting the bullet with number of entries make it a little easier to see which basket it's referring too? Not a blocker just curious

Added! Also, totally agree that it feels weird, was testing it out locally and figured if it works it works 🤷🏼

PR #348, opened 2021-02-22
LNY crawl automatically send emails

Jeremy-Xue:
Follow-up of: https://github.com/sendchinatownlove/ruby/pull/347

This will send an email whenever a new redemption is created. The language of the body is modified.

@nanxiy, right now, this will send an email for *each* redemption that is made, so if a user makes 2 redemptions at the same time, they'll receive 2 separate emails. It will be more difficult to change this to send 1 bulk email, and I don't think it will be worth it since only 3 people currently have made more than 1 redemption (likely at different times)

I'm thinking of changing the copy to say something along the lines of 'You have made an entry into [Giveaway Basket]!' instead of showing the count of entries

Old email body for reference:
Screen Shot 2021-02-21 at 9 32 58 PM

Note: This was checked out from the previous PR with the rake task, and will be rebased once that is merged in. Please ignore the first 2 commits!

### DIFF
--- a/app/controllers/redemptions_controller.rb
+++ b/app/controllers/redemptions_controller.rb
@@ -20,6 +20,8 @@ class RedemptionsController < ApplicationController
         receipts_to_redeem[i].update!(redemption: @redemption)
       end
 
+      EmailManager::LnyCrawlRedemptionSender.call(contact_id: @contact.id, redemption_id: @redemption.id)
+
       json_response(@redemption, :created)
     end
   end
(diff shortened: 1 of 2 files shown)


nanxiy:
Hm, if that's the case I'm okay with them sending 1 email for each redemption, however I'd like to keep the copy as is. I'm hoping towards the end of the crawl more people will be bulk uploading/bulk redeeming and so I think those metrics will still be valuable.

justintmckibben (approved):

jacoblb64 (approved):

PR #353, opened 2021-03-28
#770 Automated stats - Live Metrics 

stanzheng:
Automatic Metrics

for "box2" gma_count, took total gam_count base amount and added the programmatic donations from the page. 

https://trello.com/c/ix7JQibt/770-automate-stats-on-the-weebly-landing-page

image


app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,573 @@
+class StatsController < ApplicationController
+    def donation_totals
+        @result = ActiveRecord::Base.connection.execute(donation_query())
+        return @result
+      end
+
+    def gam_count
+        @GiftCardDetail = GiftCardDetail.where("single_use": true)
+        return @GiftCardDetail.length + 5140
```
jacoblb64, line 9: what's this 5140? Can we add a comment to explain?

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,573 @@
...
+    end 
+    def index
+        # some other stuff going on...
```
jacoblb64, line 12: should we remove these comments?

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,573 @@
...
+    end
+
+    def transaction_totals
+        @Item = Item.all
+        return @Item.length()
+    end
+    
+
+  def donation_query
+    query = <<-SQL
```
jacoblb64, line 29: this is a huge (kudos for working on it!) query -- any chance we can move this to a `.sql` file somewhere, then load it to a string and execute it as needed? Would be great to have intelligence while editing it and tracking it separately

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,573 @@
...
+            # pry
+            
+            box1 =  "$%s" % ActionController::Base.helpers.number_with_precision( donation_totals.getvalue(0,5), :precision => 0, :delimiter => ',')  #"$10,000"
+            box2 =  ActionController::Base.helpers.number_with_precision( gam_count, :precision => 0, :delimiter => ',') 
+            box3 = "36,573" 
+            box4 = ActionController::Base.helpers.number_with_precision( transaction_totals, :precision => 0, :delimiter => ',') 
+            box5 = sellers_total
+            box6 = "$47,689"
+            render html:  <<-HERDOC
+            <div class="" data-sr-id="3"
```
jacoblb64, line 125: Question, and maybe we can have larger offline discussion -- do you think we should be having the backend return full html, or have the front-end have it and fill in the few data fields? We have this draft PR https://github.com/sendchinatownlove/sendchinatownlove.github.io/pull/499 from Matt, not sure if you started from there.

If we do prefer to have the backend return html  -- can we have these somewhere as html files, interpolate, then return?

Also, any reason this is `show2`, what happened to `show`? ahha

jacoblb64 (approved):
Thanks for taking this on man! That is a beast of a SQL query ahha. None of my comments are blocking at all -- if this is working I'm more than happy to get it in and discuss the points later, but I do have some questions about strategy / future maintenance of this.

Also huge thank you for updating the dependencies to the latest 🙏

app/controllers/stats_controller.rb
stanzheng: this is a bit of a magic number,  the gam spreadsheet and the website shows "13818" meals added. I believe when i did the query the database only has 8000 something meals. 

So the idea of 5140 was the assumed untracked meals that were not bought directly through the GAM page so 

DB_total + magic_constant = GAM total 
8632 + 5140 = 13818 

https://docs.google.com/spreadsheets/d/1gLPrM6E6JA-pkyXGWTJzhJnR-HXydIQ95voQAxuWAGI/edit#gid=0

like for example noodlehouse was recently added so this number increased from 13318 to 13818

app/controllers/stats_controller.rb
stanzheng: > Question, and maybe we can have larger offline discussion -- do you think we should be having the backend return full html, or have the front-end have it and fill in the few data fields? We have this draft PR sendchinatownlove/sendchinatownlove.github.io#499 from Matt, not sure if you started from there.

So i thought this was also the best method, however how the app serves, getting it to render in an iframe without app headers and other attributes was very difficult. The HTML here is perfect because its a direct replication from the weebly site. 

If there was a way to hide both the app footer/header it would be easy to build a standalone component but it wasn't apparent. I pulled the latest version and ran it and it suffered from the framing issue. IE iframes only let you 100% the full view not part of it. 


> If we do prefer to have the backend return html -- can we have these somewhere as html files, interpolate, then return? 

this is fine! i ended up using HEREDOCS because i could not figure out the rails rendering engine documentation. 

I used for example 

`render stats/index` or the name of the file with an associated `view` and didn't get it working. I wasn't sure if there was a missing middleware I was missing somewhere since we aren't loading views any else in the application. 

I have no idea what happened to show1.. they must have left and never came back, ill update this to show.

stanzheng:
updated by moving the query out into separate RB files

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,58 @@
+
+# class Config
```
jacoblb64, line 2: Did you mean to leave these comments in?

jacoblb64 (approved):
Thanks for taking into account my feedback! I wish we could do `.sql` and `.html` or `.erb` files but sounds like there's some barriers to that and this is already miles better. Thanks for also explaining the source of the "magic numbers" and the philosophy of returning the iframe purely from the backend!

app/controllers/stats_controller.rb
jacoblb64: Gotchya, thanks for the explanation! I guess this is the kind of number we can eventually delegate to a Google Sheet!

app/controllers/stats_controller.rb
jacoblb64: Thanks for the in-depth answers to all of these -- sounds like there are some deeper challenges we have to overcome -- thanks for taking it as far as it is and I think we should definitely get this merged now. If we learn how to do things better or what missing middlewares could help we can always refactor!

app/controllers/stats_controller.rb
jacoblb64: Thanks for cleaning these up!

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+        # TODO(stanzheng) pull magic constant from GAM spreadsheet
+        return @GiftCardDetail.length + 5140
+    end 
+    def index
+        show(donation_totals(), sellers_total(), transaction_totals(), gam_count()) 
+    end
+
+    def sellers_total
+        @sellers = Seller.all
+        return @sellers.length()
```
justintmckibben, line 24: I think we can count this in the DB layer vs Rails layer

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,43 @@
+
+
+require_relative 'stats_helpers/donations.rb'
+require_relative 'stats_helpers/stats_html.rb'
+
+class StatsController < ApplicationController
+    def donation_totals
+        @result = ActiveRecord::Base.connection.execute($donation_query)
```
justintmckibben, line 8: Is there no way to do this via AR?

justintmckibben, line 8: nit: you can just make this whole method:

def donation_totals
        ActiveRecord::Base.connection.execute($donation_query)
end

app/controllers/stats_controller.rb
justintmckibben: `Seller.all.count`

app/controllers/stats_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    
+
+    def show(donation_totals, sellers_total, transaction_totals, gam_count)
+        box1 =  "$%s" % ActionController::Base.helpers.number_with_precision( donation_totals.getvalue(0,5), :precision => 0, :delimiter => ',')  #"$10,000"
+        box2 =  ActionController::Base.helpers.number_with_precision( gam_count, :precision => 0, :delimiter => ',') 
+        box3 = "36,573" 
+        box4 = ActionController::Base.helpers.number_with_precision( transaction_totals, :precision => 0, :delimiter => ',') 
+        box5 = sellers_total
+        box6 = "$47,689"
+        render html: stats_html(box1, box2, box3, box4, box5, box6).html_safe# render :text => @model_object.html_content
```
justintmckibben, line 40: Can we label these in the request? Maybe return a JSON with key-values so that we know what each number means from the response?

app/controllers/stats_helpers/donations.rb
```diff
@@ -0,0 +1,84 @@
+
```
justintmckibben, line 1: This is really hard to parse. If we can do this in AR it would be better — but if we have to do this via queries are there any other things we can do to construct queries vs writing them like this? At least you prepared statements? The way it's written now there is no way to recycle this

app/controllers/stats_controller.rb
stanzheng: the query is a complex copy paste of our seller payout queries, because its a query between multiple tables and entities im not sure how to express this performantly that wouldn't end up in some bottle neck or not able to be reproducable . 

https://data.heroku.com/auth/heroku/callback?code=03140850-8314-44ab-9303-3a966b46770f

app/controllers/stats_helpers/donations.rb
stanzheng: this statement is a recycle of https://data.heroku.com/dataclips/imjqsyuczmvccxmfsogcrnuzicro / templates

PR #359, opened 2021-04-11
update stats controller progress bar

ath-chen:
ok i think this should solve the nil prob!!

checked it locally with these payloads:

- "[{"sum":null,"label":"donations"}]" --> **returns 0**

- "[{"sum":null,"label":"donations"},{"sum":42000,"label":"gift_cards"},{"sum":2000,"label":"gam"}]" --> **returns 44000**


jacoblb64 (approved):
Ooh fancy, better case handling and a more elegant, less lines-of-code solution, what's not to love?

PR #362, opened 2021-04-18
add caching to seller endpoint on http and server

stanzheng:
https://trello.com/c/2Rp8Lm17/818-high-priority-investigate-caching-for-common-requests

docs 
- https://thoughtbot.com/blog/take-control-of-your-http-caching-in-rails
- https://guides.rubyonrails.org/caching_with_rails.html
- https://devcenter.heroku.com/articles/memcachier
- https://stackoverflow.com/a/56690088/2138186

### DIFF
--- a/app/controllers/sellers_controller.rb
+++ b/app/controllers/sellers_controller.rb
@@ -7,21 +7,19 @@ class SellersController < ApplicationController
   def index
     query = Validate::GetSellersQuery.new(params)
 
-    unless query.valid?
-      raise InvalidParameterError, query.errors.full_messages.to_sentence
-    end
+    raise InvalidParameterError, query.errors.full_messages.to_sentence unless query.valid?
 
-    Rails.cache.fetch("sellers", expires_in: 60.seconds) do
+    Rails.cache.fetch('sellers', expires_in: 60.seconds) do
       @sellers = Seller.order("#{query.sort_key} #{query.sort_order}")
       if stale?(@seller)
-      sellers = @sellers.map do |seller|
-      SellersHelper.generate_seller_json(
-        seller: seller
-      )
+        sellers = @sellers.map do |seller|
+          SellersHelper.generate_seller_json(
+            seller: seller
+          )
+        end
+        json_response(sellers)
       end
-      json_response(sellers)
-      end
-    end    
+    end
   end
 
   # POST /sellers
--- a/app/controllers/sellers_controller.rb
+++ b/app/controllers/sellers_controller.rb
@@ -11,14 +11,17 @@ class SellersController < ApplicationController
       raise InvalidParameterError, query.errors.full_messages.to_sentence
     end
 
-    @sellers = Seller.order("#{query.sort_key} #{query.sort_order}")
-
-    sellers = @sellers.map do |seller|
+    Rails.cache.fetch("sellers", expires_in: 60.seconds) do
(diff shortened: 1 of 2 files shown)


jacoblb64:
[[High Priority] Investigate caching for common requests](https://trello.com/c/2Rp8Lm17/818-high-priority-investigate-caching-for-common-requests)

app/controllers/sellers_controller.rb
```diff
@@ -11,14 +11,17 @@ def index
       raise InvalidParameterError, query.errors.full_messages.to_sentence
     end
 
-    @sellers = Seller.order("#{query.sort_key} #{query.sort_order}")
-
-    sellers = @sellers.map do |seller|
+    Rails.cache.fetch("sellers", expires_in: 60.seconds) do
```
jacoblb64, line 14: There was some offline discussion about this -- but I think we can _way_ up this expiry time -- 60s solves for the forward / back case, but I think 1hr (or even 2hrs) is even better to just overall speed up most page loads.

Eventually I think we could even create a background job to keep it fresh and spare the person who gets the cache miss

app/controllers/sellers_controller.rb
```diff
@@ -11,14 +11,17 @@ def index
...
+      @sellers = Seller.order("#{query.sort_key} #{query.sort_order}")
+      if stale?(@seller)
+      sellers = @sellers.map do |seller|
```
jacoblb64, line 17: nit: spacing? you can run `rubocop -a path/to/this/file` to get the autoformatting

config/environments/production.rb
```diff
@@ -79,6 +80,19 @@
 
   # Use a different cache store in production.
   # config.cache_store = :mem_cache_store
+  config.action_controller.perform_caching = true
+
+  config.cache_store = :mem_cache_store,
+                       (ENV['MEMCACHIER_SERVERS'] || '').split(','),
```
jacoblb64, line 86: Would be curious how to test this locally / what happens if these aren't defined. Also, what setup did you / do we have to do in heroku for staging / prod?

jacoblb64 (approved):
Looks great man! This is definitely one of those things that a lot of reading, research, and offline setup manifests as a few lines of code 😬 , but I really appreciate you taking care if it -- and it looks like overall the setup will help us cache other low-hanging fruit as well! (looking at you progress bar and GAM page 👀 ).

Would love to get Rohan's take and see what he thinks as well.

Also, was Memchachier easier out-of-the-box on heroku than redis or other solutions?

config/environments/production.rb
stanzheng: in development i dont think its enabled at all

app/controllers/sellers_controller.rb
stanzheng: will this cause donation totals to be off?

jacoblb64: It will cause them to be off, or at least "lag" behind what's really happening. But for the sake of performance, Ling, Nanxi, and Brian all signed off I believe.

config/environments/production.rb
jacoblb64: Gotchya, that's probably for the best!

app/controllers/sellers_controller.rb
stanzheng: increased to 1 hour

PR #367, opened 2021-04-24
Wilson/add id for delivery option

wilsonj806:
# [Ticket](https://trello.com/c/Bs9FTAhG)
Also this PR does some other stuff as well, see list:
- adds a migration to add a `delivery_type_id` column to delivery types with an id field that's similar to how `sellers.seller_id` is formatted
- adds a rake task to add delivery types

# Breaking Changes
Will need to run `bundle exec rake one_time_tasks:build_delivery_types` on successful deploy

Also rollbacked the migration that made `campgians.distributor_id` not null since Mega-GAMS don't use it

# Updated Request
Post requests now look like this:
**Endpoint**:  `POST /sellers/seller_id/delivery_options`
JSON:
```json
{
   "url":"google.com",
   "delivery_type_id": "chowbus"
}
```

Where delivery_types.delivery_type_id emulates how sellers.seller_id looks like. See `lib/tasks/one_time_tasks.rake` for list of delivery_type_ids.


jacoblb64 (approved):
Looks all great to me, thanks for taking care of this!

PR #374, opened 2021-05-06
Stanleyz/livestats

stanzheng:
- use an example spreadsheet here to translate values excluding seller totals


app/controllers/stats_controller.rb
```diff
@@ -21,41 +23,47 @@ def sellers_total
...
   def transaction_totals
     Item.all.count
   end
 
   def show(donation_totals, sellers_total, transaction_totals, gam_count)
-    donation_totals = '$%s' % ActionController::Base.helpers.number_with_precision(donation_totals, precision: 0, delimiter: ',')  # "$10,000"
+    outside_db_gam_contributions = 0
+    outside_db_fundaising_contributions = 0
+    outside_db_fundaising_foodcrawl_raised = 36, 573
+    outside_db_luc_raised = 47, 689
```
jacoblb64, line 34: these commas make me nervous, maybe `rubocop -a` these? I think the ruby way is `47_689`

app/helpers/spreadsheet.rb
```diff
@@ -0,0 +1,41 @@
...
+# First worksheet of
+# https://docs.google.com/spreadsheet/ccc?key=pz7XtlQC-PYx-jrVMJErTcg
+# Or https://docs.google.com/a/someone.com/spreadsheets/d/pz7XtlQC-PYx-jrVMJErTcg/edit?usp=drive_web
+# ws = session.file_by_id("1RFbedwlWR2I0d0ELllCeha9QCdZK_PQyJt_RcaerexU").worksheets[0]
+
+
+
+module LiveStats
+  def self.pull
+    $google_config = ENV['GOOGLEDRIVE_SECRET']
```
jacoblb64, line 20: is this the same kind of credentials as used by GCP? If so I've been using `GOOGLE_APPLICATION_CREDENTIALS` which the GCP plugin expects

app/controllers/stats_controller.rb
```diff
@@ -21,41 +23,47 @@ def sellers_total
...
-    foodcrawl_raised    = '36,573'
-    transaction_totals  = ActionController::Base.helpers.number_with_precision(transaction_totals,
-                                                                               precision: 0, delimiter: ',')
-    sellers_total
-    luc_raised = '$47,689'
+    # box3
+    foodcrawl_raised    = outside_db_fundaising_foodcrawl_raised
+
+    # box4
+    total_vouchers_giftcards = ActionController::Base.helpers.number_with_precision(transaction_totals,
```
jacoblb64, line 60: should we do total items here -- or switch it to total payment_intents? there was some discussion of this with Ling and co, maybe we should take this to discord?

jacoblb64 (approved):
This looks awesome, great work man! Left some comments but non-blocking

PR #383, opened 2021-07-30
Fix email disaster

jacoblb64:
- add `CORRECTED:` to the subject of all previous emails we're re-sending to
- fix email name bug, but as a fallback call folks 'SCL Supporter' in the worst case 😂 
- adding a try/catch to skip instead of cancelling the whole process


wilsonj806 (approved):
Looks good although I don't know the full context!

stanzheng:
LGTM

